Issue #5, opened 2019-02-22
AmqpConnection CreateSubscriber should not assert exchange


WonderPanda:
Currently the `createSubscriber method in `AmqpConnection` which is used by the `RpcSubscribe` decorator attempts to assert the exchange before subscribing. As part of this it's passing through a default exchangeType value for `topic`.

This results in errors at startups exchange types other than topic:

```
RabbitMQModule.build({
      exchanges: [
        {
          name: 'exchange1',
          type: 'topic',
        },
        {
          name: 'exchange2',
          type: 'fanout',
        },
      ],
      uri: 'amqp://rabbitmq:rabbitmq@localhost:5672',
    })
```

produces the error:

`Error: Channel closed by server: 406 (PRECONDITION-FAILED) with message "PRECONDITION_FAILED - inequivalent arg 'type' for exchange 'exchange2' in vhost '/': received 'topic' but current is 'fanout'"`

This should be handled by the module init based on the exchanges provided to the RabbitMQModule so exchanged can be asserted once (with proper values) instead of allowing it to fall through to individual handlers.

Libraries that choose to consume AmqpConnection for use cases outside of the Decorator patterns exposed by the module will be responsible for doing their own exchange assertion.

WonderPanda:
Fixed in @nestjs-plus/rabbitmq:0.2.1

PR #9, opened 2019-02-22
Remove responsibility of asserting exchanges from rpc/subscribers

WonderPanda:
- Extend exchange config to allow amqplib options
- Extend public API with index.tx exports
- Allowed setting the default exchange type externally


WonderPanda:
Fixes #5

PR #18, opened 2019-03-06
Fix for Providers that don't provide classType

WonderPanda:
- Updated the API to differentiate between the constructor of the
injectable vs the constructor of the dependency

- Add better checking and reflection to make sure we don't reflect
on undefined objects

- Addresses https://github.com/WonderPanda/nestjs-plus/issues/17


WonderPanda:
Addresses https://github.com/WonderPanda/nestjs-plus/issues/17

Issue #26, opened 2019-03-19
Upgrade Core NestJS Dependencies to Nest 6


WonderPanda:
Now that Nest 6 is available, the @nestjs-plus packages need to be updated to support it. This will entail a new major version release for all current packages which means that version 5 support will no longer receive new functionality (to avoid semver issues).

The discovery package is the biggest blocker now as it depends on NestJS DI internals which have been updated significantly.

WonderPanda:
This may be blocked by https://github.com/nestjs/nest/issues/1786 as it appears there's some core functionality around modules and dependency injection that isn't working properly. It's unclear whether this is only in the context of TestingModules or whether it extends further

PR #27, opened 2019-03-24
feat(all packages): upgrade to Nest v6

WonderPanda:
BREAKING CHANGE: upgrades underlying nest package dependencies to v6 versions


WonderPanda:
Closes #26

PR #57, opened 2019-10-28
fix(rabbitmq): fix mq message type not json

Dn9x:
mq接收的消息不一定是json字符串，可能是任意的编码数据，如果直接json转码，一定会报错。
建议转码部分交由业务处理，框架部分只是封装抽象业务，数据进行透明传递。

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -113,8 +113,14 @@ export class AmqpConnection {
         if (msg == null) {
           throw new Error('Received null message');
         }
-
-        const message = JSON.parse(msg.content.toString()) as T;
+        let message = {} as T;
+        try {
+          message = JSON.parse(msg.content.toString()) as T;
+        } catch (e) {
+          // 这里可能不是json格式，比如protobuf格式
+          // 先直接使用try进行捕获，合理的方案应该是直接在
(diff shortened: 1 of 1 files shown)


WonderPanda:
Hey @Dn9x Sorry, I'm just seeing this now. It seems reasonable but I'll need to translate the stuff you've provided

WonderPanda:
Translation:

The message received by mq is not necessarily a json string, it may be any encoded data. If you directly transcode json, it will report an error.
It is recommended that the transcoding part be handed over to the business, and the framework part only encapsulates the abstract business and the data is transparently transmitted.

WonderPanda:
@Dn9x
感谢您的贡献。我认为您绝对正确的，该框架不应对通过RabbitMQ发送的消息数据的性质进行假设。目前，这是一个不错的解决方法，但是我将考虑如何在将来的版本中使其更加灵活。

Thank you for your contribution. I think that you are absolutely correct that the framework should not make assumptions about the nature of the message data being sent over RabbitMQ. This is a good workaround for right now but I will consider how we can make this more flexible in future releases.

WonderPanda (approved):

WonderPanda (approved):

WonderPanda:
@Dn9x 
您的更改已发布，并且可以在@ nestjs-plus / rabbitmq @ 1.4.2中找到

your changes have been published and are available in @nestjs-plus/rabbitmq@1.4.2

Issue #71, opened 2019-11-10
Expose PublishOptions on Request


elucidsoft:
Need to expose PublishOptions to Request so that you can modify them if necessary.  It's an optional parameter so it should not be a breaking change.

PR #72, opened 2019-11-10
Expose publish options

elucidsoft:
This resolves PR #71


elucidsoft:
Fixing build...

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -77,7 +78,8 @@ export class AmqpConnection {
 
     this.publish(requestOptions.exchange, requestOptions.routingKey, payload, {
       replyTo: DIRECT_REPLY_QUEUE,
-      correlationId
+      correlationId,
+      ...publishOptions
```
WonderPanda, line 82: The signature for `publish` is:
```
public publish(
    exchange: string,
    routingKey: string,
    message: any,
    options?: amqplib.Options.Publish
  )
```
So it's expecting the options as a 4th arg.  Looks like you're spreading them into the 3rd (message) arg here so I don't think it'll interact properly with the underlying amqplib

WonderPanda (changes requested):
This repo depends on using yarn to properly leverage yarn workspaces + lerna to support monorepo publishing. So there shouldn't be a package-lock.json included

packages/rabbitmq/src/amqp/connection.ts
elucidsoft: let me double check I did this right...

WonderPanda: Whoops nevermind I misread this

elucidsoft: Yes, I wanted to keep the values that were specific to request but allow additional items to be added.

elucidsoft:
I would like to abandon this...

PR #80, opened 2019-12-10
Support connection resiliency

azuker:
First, I'd like to say that I love this package and I've been using it for quite a while.
I was really missing a feature of having connection resiliency, and this is what this PR is about, hopefully it should meet your vision.

**Key achievements:**
- Allow bootstrapping the app without a connection being available
- Allow disconnections on the fly and try to reconnect seamlessly
- Setup all the related RPC and subscriber handlers whenever a connection is established
- Do the above by leveraging `amqp-connection-manager`

**Includes the following as well:**
- Changes to rabbitmq `README.md` to describe the connection related behavior
- Added relevant parameters to `RabbitMQConfig`
- Aligned tests

**Versioning:**
Currently, the implementation mimics the same previous behavior by default, meaning the app crashes when a connection cannot be made.
In order to enable the new mode, the developer needs to configure it explicitly - `connectionInitOptions: { wait: false }`
We should consider changing the default in the next major release.

Note that I didn't change anything in regards to package versions.


WonderPanda:
@azuker Hey! Thanks for taking the time to put together a great addition to the library. I've been wanting to improve connection resiliency for a while now but life gets in the way so I really appreciate this. Been a bit busy so just getting to this now.

This is the first time I've had a sizeable contribution from anyone with potentially breaking changes so I want to take a little bit to look through

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -23,43 +25,133 @@ const defaultConfig = {
...
-
-    await Promise.all(
-      this.config.exchanges.map(async x =>
-        this._channel.assertExchange(
-          x.name,
-          x.type || this.config.defaultExchangeType,
-          x.options
-        )
+  public async init(): Promise<void> {
+    const {
```
WonderPanda: This should be moved up and included with `connectionInit` in the `defaultConfig` object. That way anyone looking at this file at a glance can see what the default options that are provided are

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -37,14 +38,22 @@ export interface MessageHandlerOptions {
...
-  uri: string;
+  uri: string | string[];
   prefetchCount?: number;
   exchanges?: RabbitMQExchangeConfig[];
   defaultRpcTimeout?: number;
   defaultExchangeType?: string;
   defaultRpcErrorBehavior?: MessageHandlerErrorBehavior;
   defaultSubscribeErrorBehavior?: MessageHandlerErrorBehavior;
+  connectionInit?: ConnectionInitOptions;
+  connectionManager?: amqpConnectionManager.AmqpConnectionManagerOptions;
```
WonderPanda: Can these properties both be renamed to have the `Options` suffix. I found the PR a bit more confusing at first glance because I missed the type on `connectionManager` and mistook it as a place where you would pass in your own pre-instantiated `connectionManager` instance.

`connectionManagerOptions` makes it explicit from a readability standpoint

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -23,43 +25,133 @@ const defaultConfig = {
...
 
 export class AmqpConnection {
-  private messageSubject = new Subject<CorrelationMessage>();
-  private _connection!: amqplib.Connection;
-  private _channel!: amqplib.Channel;
-  private config: Required<RabbitMQConfig>;
+  private readonly messageSubject = new Subject<CorrelationMessage>();
+  private readonly config: Required<RabbitMQConfig>;
+  private readonly logger: Logger;
+  private _connection!: amqpcon.AmqpConnectionManager;
```
WonderPanda: From a naming standpoint, I'd like to reverse this so that we're introducing new properties. I think using `managedConnection` and `managedChannel` would make sense. Then we could include a deprecation warning log in the old properties in their `getters` so that consumers of the library have time to switch over instead of introducing a breaking package change.

If we do that I think we can release this entire change under a minor version bump that will be backwards compatible

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -23,43 +25,133 @@ const defaultConfig = {
...
+    this._connection.on('connect', () => {
+      this.logger.log('Successfully connected to a RabbitMQ broker');
+    });
+
+    this._channel = this._connection.createChannel({
+      name: AmqpConnection.name
+    });
+
+    if (firstInitCallback) {
+      this._channel.once('connect', () => firstInitCallback());
```
WonderPanda: Is there a specific use case for having a callback for the first connection? It seems like it would be better to have an `onConnection` property which could be `(connectionCount: number) => void` and we could track the number of times a connection has been established as a property of this class. Then a consumer of the library can attach a much more generic callback with more extensible options for processing events

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -23,43 +25,133 @@ const defaultConfig = {
...
+          err.message
+        } ${err.stack}`
+      );
+    });
+    this._channel.on('close', () => {
+      this._rawChannel = undefined;
+      this.logger.log('Successfully closed a RabbitMQ channel');
+    });
+
+    await this._channel.addSetup(this.setupInitChannel.bind(this));
```
WonderPanda: I'd rather have an arrow function here instead of doing `.bind(this)` it's a bit of personal preference but I don't like having to spend cycles worrying about `this` context with JS

WonderPanda (changes requested):

WonderPanda:
Would love if this could be rebased against the latest `origin/develop` as well

azuker:
Thanks for the review, I'm glad to get a chance to contribute.
I'm currently abroad I'll change and resolve the comments when I return later this week.

packages/rabbitmq/src/amqp/connection.ts
azuker: I needed that for the purpose of the tests.
Before, the connection module finished bootstrapping after the connection had been made or failed altogether. Therefore, the tests ran after a connection was available and passed successfully.

Now, since the tests had to "wait" for a connection to be made, I made the "init" function that returns a promise and relevant config options to configure it to "wait" mode.
To make it promise-able, I passed down the resolve callback.

I like your approach and we could switch to that and have the tests use it. However, currently this implementation supports the "old" way of having it fail if no connection is available without the developer having to do anything.

I'll add the callback you mentioned, but do you still want to keep this option of wait-and-reject-timeout too?
Also, do you want that property to be a simple callback or would you like to utilize an event-ingn approach of some sort? (EventEmitter / RxJs / etc.)

BTW - I think I still have a problem with this that I need to fix. The tests failed if.I didn't create an exchange manually beforehand, and I suppose it is because the promise is resolved before it asserts the exchanges - I plan to change that to be after

azuker: I'll make it simpler. I'll keep the wait+reject mode but I will do that using RxJs observables following your idea.

packages/rabbitmq/src/amqp/connection.ts
azuker: Done

packages/rabbitmq/src/rabbitmq.interfaces.ts
azuker: Done

packages/rabbitmq/src/amqp/connection.ts
azuker: Done
I didn't add the deprecation warning but it's now actually backward compatible - the default mimics the previous behavior. (I'll explain in a general comment above)

packages/rabbitmq/src/amqp/connection.ts
azuker: Done

packages/rabbitmq/src/amqp/connection.ts
azuker: Done

azuker:
I changed the implementation to reflect the same previous behavior by default.
This means that these changes should be backward compatible now.

Now the developer needs to configure it to enable the mode of having the application launch properly when no connection can be made

azuker:
If you like me to open the PR against the origin/develop instead then do let me know

WonderPanda:
Thanks for pushing these changes @azuker! I have a busy work day today and a Christmas party tonight so I won't have time to fully test out all the additions but I will try and get a release out that includes connection resiliency by this weekend.

azuker:
Sure thing, that's actually very quick :) thanks!
Have fun and merry Christmas (to come)!

azuker:
Sorry for pushing another set of changes after requesting a review just now.
I had to do it to fix the following -
- Fix use of `isConnected` to `isConnected()`
- Mimic the previous behavior better to be backward-compatible - I no longer reset the channel to undefined, so when there is no connection there will still be errors when using the channel instead of when accessing the getter.

azuker:
There's another thing we need to consider.
Currently, the prefetch is part of the channel setup, meaning it should take place every time a connection is made. We might want to change it to do it only once,

azuker:
Jesse, do you know when you'll be able to get back to this PR anytime soon?
I'm simply wondering because we need these changes for a project I'm leading. thanks!

WonderPanda:
@azuker Yes, my apologies I overestimated how much time I'd have over the holidays for OSS stuff. I promise that I will get a new version out with these additions this week

WonderPanda:
Hey @azuker currently looking at this, great work! I'm going to rebase and do a little bit of tweaks and clean up to get this merged in. It will be easier for me to open up a new PR by pulling the branch from your fork into this repo. I may clean up some commit messages as well so it better lines up for changelog generation but you'll still receive attribution for the work on your commits.

One thing I think is missing is that we currently don't expose a top level way to nicely send messages now. The old calls to `publish` now fail (as expected) if they're called before things have been fully setup. Looking at the features of `amqp-connection-manager` I think we want to leverage their `sendToQueue` functionality directly on the `AmqpConnection` object that is exposed from `golevelup`. I like the idea of it keeping them in memory and sending them automatically once connection is established.

Working through things now I may update the comments a bit more as I go. Do you use Discord at all? I'd be happy to collaborate some time in a more real time format depending on your timezone

WonderPanda:
I'm just going through some testing scenarios and am seeing something that doesn't feel right to me. I grabbed a random e2e test: https://github.com/golevelup/nestjs/blob/master/integration/rabbitmq/e2e/subscribe.e2e-spec.ts
and am purposely delaying RabbitMQ from starting to simulate it coming available later after the app is bootstrapped.

When I use the `amqpConnection.managedChannel.publish` which queues the message and sends it when the connection comes online seems to fire but the `testHandler` jest function is never invoked. This seems like a problem to me as it might mean that the Subscriber logic isn't properly being attached in the case when the connection wasn't immediately available. I'll dig into it some more and try to figure out what's up

azuker:
I live in Israel so timezone is GMT+2.
Live collaboration is a nice idea though I don't currently use Discord, we can try it out.

Regarding the above comments -

1) **Cleaning up and new PR**:
Sounds good, especially if it preserves my record of contribution as well :)

2) **publish**
I think it's a nice touch too although it's not necessarily what all implementations want.
Sometimes you want it to fail and not hang forever.
Plus, this is a behavioral breaking change and should require bumping a major version.

I recommend we keep it as it is now to preserve backward compatibility and design a new API for allowing both options afterwards.

3) **`amqpConnection.managedCjhannel.publish` test**
I'm not sure I understand the test scenario.
If you like, you should be able to edit my PR, so you can push the test in and I can have a look as well.

**Side Note**
I have additional plans to contribute if you're willing to accept, for example -

- Allow configuring nest context type (currently it's http so there is no way to know in guards or interceptors if it reached from actual http or through this transport)
- Use RpcContextCreator and support pipes, or better yet, implement a custom microservice transport
- Support not hard-coded exchange and queue names via presets (the decorators will not receive the exchange name but rather a preset name that is configured in the module forRoot)
- Research if having a channel per consume is better in terms of reliability and acks, then implement if so

WonderPanda:
@azuker Replaced by https://github.com/golevelup/nestjs/pull/94

WonderPanda:
This has been published and is now available:

```
Successfully published:
 - @golevelup/nestjs-rabbitmq@1.9.0
```

I definitely want to revisit some things related to the `amqp-connection-manager` library but am very much interested in the additional functionality that you are thinking about.

If you're interested in chatting a bit about it I'd be happy to communicate in a more real time manner to collaborate on future advancements. I'm on the NestJS Discord Server as WonderPanda. There are a few other members there who I collaborate with would be awesome to get your insight in some conversations.

If Discord isn't your thing what messaging platforms so you tend to use?

azuker:
Terrific! We're going to use it right away in one of my projects, thanks!

To begin with, is it okay if I email to your profile's email address so we can communicate outside the PR?
Personally, I use zoom/skype/teams and VS Code Live Share for collaborating when using VS Code.
Nevertheless, I don't mind giving Discord a try, always open for new things :)

WonderPanda:
@azuker Yeah, feel free to email at the address on my profile! I'm excited that you'll be immediately making use of some of the new functionality and would love to collaborate on stuff for the future

PR #82, opened 2019-12-12
fix(discovery): discovery of controllers/providers in lazy way

### LINKED ISSUE #81: instance properties of discovered controller are undefined
Hi!

Have found such a strange issue: instance properties of discovered controllers are `undefined` in discovery results.

I think that it can happen because controllers discovery is in the discovery service constructor, so it should be delayed at least to the finish of the app bootstraping process. Moreover, discovery process could be implemented in lazy way.

I'll prepare PR for that shortly.

Thanks.

zarv1k:
Fixes #81

### DIFF
--- a/packages/discovery/src/discovery.service.ts
+++ b/packages/discovery/src/discovery.service.ts
@@ -54,39 +54,22 @@ export const withMetaAtKey: (
 
 @Injectable()
 export class DiscoveryService {
-  private readonly discoveredControllers: Promise<DiscoveredClass[]>;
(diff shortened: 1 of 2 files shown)


zarv1k:
@WonderPanda could you review please?

WonderPanda:
@zarv1k I've been busy with some holiday stuff but I'm excited to see your contribution. I will take look now and see about getting it released

WonderPanda (approved):

PR #90, opened 2020-01-02
nestjs-discovery package.json update

robertpallas:
When `lodash` is not dependency of the Nest based app production installed start fails as `lodash` is not found (`Cannot find module 'lodash'`)

 - add `lodash` as dependency
 - update bugs, repo & homepage urls

### DIFF
--- a/packages/discovery/package.json
+++ b/packages/discovery/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@golevelup/nestjs-discovery",
-  "version": "2.2.1",
+  "version": "2.2.2",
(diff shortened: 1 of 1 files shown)


WonderPanda:
Hey @robertpallas thanks for the contribution!
I use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) specification for this repo so that I can generate change logs and version bumps automatically. Because your commit message doesn't follow the spec I rebased and opened a new PR with an amended commit message containing your changes:

https://github.com/golevelup/nestjs/pull/91

PR #91, opened 2020-01-06
fix(discovery): add lodash as dependency

WonderPanda:
When `lodash` is not dependency of the Nest based app production installed start fails as `lodash` is not found (`Cannot find module 'lodash'`)

 - add `lodash` as dependency
 - update bugs, repo & homepage urls


WonderPanda:
@robertpallas new package versions have been published with your change:

 - @golevelup/nestjs-discovery@2.2.2
 - @golevelup/nestjs-rabbitmq@1.8.1

PR #92, opened 2020-01-09
style(eslint): add eslint and sonarcloud

jmcdo29:
EsLint configured with sonarjs, typescript, and prettier. Sonarcloud property file also added for
sonarcloud integration

Should be no breaking changes, all tests still pass with no problems


packages/rabbitmq/src/rabbitmq.module.ts
```diff
@@ -125,5 +125,6 @@ export class RabbitMQModule
         })
       );
     }
+    32;
```
jmcdo29: Huh, somehow missed this. I'll remove it and update the PR

PR #99, opened 2020-01-21
feat(rabbitmq): enable conditional rpc/subscribe handler registration

### LINKED ISSUE #98: Support for disabling consumers using an application config
I have many instances of my application running, some of which I'd like to consume from RabbitMQ queues and others that should be dedicated to serving API requests.

I'm using the RabbitSubscribe decorator on a service method to connect a consumer.
```
@RabbitSubscribe({
    exchange: 'exampleExchange.created',
    routingKey: 'exampleExchange.*',
    queue: 'exampleExchange.created',
  })
  public async process(event): Promise<void> {
    ...
  }
```
Is there a way to stop this from consuming from the queue based on my application config, while allowing the application to publish to the same exchanges/queues?

WonderPanda:
fix #98


WonderPanda:
Ignoring SonarCloud gate on this one because I'm fine with the duplication that these tests introduced

PR #121, opened 2020-03-28
feat(common): injectable decorator factory

### LINKED ISSUE #120: Create Utility in common to create injectable token decorators
Instead of having to do `@Inject(SOME_TOKEN_KEY_THING)` all over the place it's better to create a decorator factory such that you can easily create custom decorators which kid the token. Eg. `@InjectThing()`

WonderPanda:
fix #120


packages/common/src/injectDecoratorFactory.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Instead of using @Inject(SOME_THING_TOKEN) this can be used to create a new named Decorator
+ * such as @InjectSomeThing() which will hide the token details from users making APIs easier
+ * to consume
+ * @param token
+ */
+export const makeInjectableDecorator = (
+  token: string | symbol
+): (() => ParameterDecorator) => () => (target, key, descriptor) =>
+  Inject(token)(target, key, descriptor);
```
jmcdo29, line 14: Instead of having to say `Inject(token)(target, key, descriptor)`, couldn't you just return the `Inject(token)` and let typescript take care of the rest? Overall making the function
```ts
export const makeInjectableDecorator = (token: string | symbol) => Inject(token);
```

WonderPanda: I don't think so. That would return the result of calling Inject instead of a function that calls Inject on the thing that it's applied to

jmcdo29: Oh, you know what, you're right. I was thinking in the context of actually having `makeInjectableDecorator` the decorator, which absolutely doesn't make sense. 

```ts
export function OgmaLogger(topic: string | Function) {
  return Inject(
    createProviderToken(typeof topic === 'function' ? topic.name : topic),
  );
}
```
This is the example I was thinking of, but that's not quite a dynamic use case

WonderPanda: Technically it could be `export const makeInjectableDecorator = (token: string | symbol) => () =>
  Inject(token);` instead (note the extra arrow) but then the typing no longer aligns with `ParameterDecorator`

PR #126, opened 2020-04-02
feat(rabbitmq): error behaviour for replying error

perf2711:
I need my RPC service to return me errors, in case something goes wrong. Thus, a new error behavior has been added: 
```
  /**
   * If an exception occurs while handling the message, the error will be serialized and published on the `replyTo` queue.
   * If `replyTo` is not provided, the message will be NACKed without requeueing.
   * If publish fails, message will be NACKed and requeued.
   */
  REPLYERRORANDACK,
```

If this is set, the handler will behave as it is described in the comment.

Additionaly, I added a `try/catch` on JSON message parsing, and allowed `undefined` messages to go into the service. I believe that the service should be able to determine what to do with this. As of before, there was no option to publish an empty message to the service (`JSON.parse` would throw an error). This was the push for me to create the error behavior, as the service would by default endlessly requeue the message without me knowing what is going on.

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -233,7 +233,7 @@ export class AmqpConnection {
     msgOptions: MessageHandlerOptions,
     channel: amqplib.ConfirmChannel
   ): Promise<void> {
-    const { exchange, routingKey } = msgOptions;
+    const { exchange, routingKey, allowNonJsonMessages } = msgOptions;
 
     const { queue } = await channel.assertQueue(
       msgOptions.queue || '',
@@ -252,7 +252,12 @@ export class AmqpConnection {
           throw new Error('Received null message');
         }
 
-        const response = await this.handleMessage(handler, msg);
+        const response = await this.handleMessage(
+          handler,
+          msg,
+          allowNonJsonMessages
+        );
+
         if (response instanceof Nack) {
           channel.nack(msg, false, response.requeue);
           return;
@@ -279,10 +284,6 @@ export class AmqpConnection {
             }
             case MessageHandlerErrorBehavior.REQUEUE: {
               channel.nack(msg, false, true);
-              break;
-            }
-            case MessageHandlerErrorBehavior.REPLYERRORANDACK: {
-              this.handleReplyAndAckError(channel, msg, e);
               break;
             }
             default:
@@ -313,7 +314,7 @@ export class AmqpConnection {
     rpcOptions: MessageHandlerOptions,
     channel: amqplib.ConfirmChannel
   ) {
-    const { exchange, routingKey } = rpcOptions;
+    const { exchange, routingKey, allowNonJsonMessages } = rpcOptions;
 
     const { queue } = await channel.assertQueue(
       rpcOptions.queue || '',
@@ -332,7 +333,12 @@ export class AmqpConnection {
           throw new Error('Received null message');
         }
 
-        const response = await this.handleMessage(handler, msg);
+        const response = await this.handleMessage(
+          handler,
+          msg,
+          allowNonJsonMessages
+        );
+
         if (response instanceof Nack) {
           channel.nack(msg, false, response.requeue);
           return;
@@ -381,12 +387,18 @@ export class AmqpConnection {
       throw new Error('AMQP connection is not available');
     }
 
-    this._channel.publish(
-      exchange,
-      routingKey,
-      Buffer.from(JSON.stringify(message)),
-      options
-    );
+    let buffer: Buffer;
+    if (message instanceof Buffer) {
+      buffer = message;
+    } else if (message instanceof Uint8Array) {
+      buffer = Buffer.from(message);
+    } else if (message != null) {
+      buffer = Buffer.from(JSON.stringify(message));
+    } else {
+      buffer = Buffer.alloc(0);
+    }
+
+    this._channel.publish(exchange, routingKey, buffer, options);
   }
 
   private handleReplyAndAckError(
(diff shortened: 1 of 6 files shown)


Arebisc (commented):
Looks fine. Thanks @perf2711

WonderPanda:
Hey @perf2711 thanks for this contribution. One thing that I would definitely like to see before this could be merged is some additions to the integration test suites that validate the expected behavior. Let me know if you need some help getting pointed in the right direction

perf2711:
@WonderPanda sure, I'll try to write some tests for this.
To be sure, you mean the e2e tests in `integration/rabbitmq/e2e`?

WonderPanda:
@perf2711 Yes, you can use the included docker-compose file to get a running instance of RabbitMQ and then use `yarn test:integration`

WonderPanda:
Hey @perf2711 I've been looking at this a little bit more and I can understand your desire for this solution but I think it might be possible to implement the ability for you to handle the errors in this way in a more generic fashion. Instead of adding a new error handling behavior at the connection level what if we added the option to supply your own error handling logic at the decorator level so that people can flexibly handle error cases differently across different queues/exchanges etc.
I'm wondering about adding in a new option method to the decorator that might look something like this:

```
errorHandler?: (
      connection: AmqpConnection,
      channel: amqplib.Channel,
      msg: amqplib.ConsumeMessage
    ) => void | Promise<void> 
```

That way if an error happens and an errorHandler was provided the current connection, channel and message that caused the error will be provided to your error handler which can do whatever it wants with this information. That would allow the community to supply additional plugins that cover difference scenarios and desired behavior without needing to bloat the core AmqpConnection code with more and more Error Handling Behaviors.

Tracking this as an enhancement issue here #127

perf2711:
Hmm, that is a neat idea. To be honest, I don't know the NestJS internals that much, but could we extend it to guards, or injectable handlers? Looking to it now.

perf2711:
By the way, I pushed some integration tests to the previous feature. I don't mind them being changed, if we go further with error handlers 😊

WonderPanda:
Hey @perf2711 sorry for the delay on this, I've had a few other open source packages that I'm working on implementing and publishing from the monorepo so this has been on the backburner for me. I would really like to go with the more flexible approach so that the community can plugin their own customized error logic as opposed to baking in more named error handling strategies

perf2711:
@WonderPanda I added the callback functionality. See if it complies with your idea 😊

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -376,11 +363,75 @@ export class AmqpConnection {
...
+    if (msg == null) {
+      return;
+    } else {
+      try {
+        if (msgOptions.errorCallbacks) {
+          for (const callback of msgOptions.errorCallbacks) {
+            await callback(channel, msg, error);
+          }
+        }
+      } finally {
```
WonderPanda, line 421: I feel like at this point we should think about removing this finally case and replacing the switch case with new named Error Handling Functions that represent the old behaviors, similar to the ReplyErrorCallback that you made. If we run the `msgOptions.errorCallbacks` as well as one from the legacy behaviors here we we will likely try to ack or nack a message that has already been acked or nacked which could lead to more problems

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -45,8 +45,16 @@ export interface MessageHandlerOptions {
   queue?: string;
   queueOptions?: QueueOptions;
   errorBehavior?: MessageHandlerErrorBehavior;
+  errorCallbacks?: IMessageErrorCallback[];
+  allowNonJsonMessages?: boolean;
```
WonderPanda, line 49: I like the intent for `allowNonJsonMessages` but I think that it doesn't belong. One of the follow up things I'd like to do with the RabbitMQ library is provide pluggable serialization/deserialization of messages instead of assuming JSON on the users's behalf. For now I'd rather just let the JSON parsing throw an error and leave it to the errorCallback to decide what to do with the error

WonderPanda (changes requested):

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -45,8 +45,16 @@ export interface MessageHandlerOptions {
...
 }
 
+export type IMessageErrorCallback = (
```
WonderPanda, line 52: The signature should be this:
```
IMessageErrorCallback = (
      connection: AmqpConnection,
      channel: amqplib.Channel,
      msg: amqplib.ConsumeMessage,
      error: any
    ) => void | Promise<void> 
```
so that the user has access to the AmqpConnection instance when they're implementing error handling. That way they can optionally publish messages or do other more advanced stuff.

I think we should restrict return to `void | Promise<void>` to prevent them from trying to return a value from the error handling function because nothing would be done with the return object

WonderPanda:
@perf2711 Thanks for your persistence with this! I'm starting to get excited about the direction we're taking things the API is really shaping up. I left some feedback for you to check out, hopefully it's not annoying I know I'm picky sometimes

I really appreciate you taking the time to help put together a significant contribution for the library

packages/rabbitmq/src/amqp/connection.ts
perf2711: Should we catch errors in handlers and continue other handler execution if one fails?

packages/rabbitmq/src/rabbitmq.interfaces.ts
perf2711: All right, but we should do something about empty messages. At this point, sending empty (i.e. empty byte array) to the queue will result in an JSON parsing error.

WonderPanda:
Hey @perf2711 I pulled your work into a new branch so I could get it rebased on top of master and to add a few cleanup items. Closing this in favor of #134 there are a few small changes/additions to the work that you did but you'll still receive credit for all the contributions you've included this far

PR #134, opened 2020-04-16
Feature/rabbitmq error handling

WonderPanda:


WonderPanda:
@perf2711 I've decided to change the signature so that only a single error handler can be applied instead of an array. It introduces too many weird edge cases where the first handler could ack or nack the message and then a subsequent one might try to do something different.

The API is highly composable now so you should be able to achieve any kind of logic by wrapping around or extending the included error handlers. I've also deprecated the old error handling behavior enum and replaced each of them with an equivalent errorHandler

WonderPanda:
@perf2711 The other thing I noticed is that when writing custom error handlers it's important that you ack or nack the message manually as this used to be guaranteed with the old behavior. In https://github.com/golevelup/nestjs/pull/134/files#diff-e2e02a1faf5f35c485133e73ca87f588 with the custom behavior that you added for the new test suite I needed to add the additional call to `ack` to prevent tests from failing on subsequent runs because the messages would be piling up in the queue and never being removed

WonderPanda:
@perf2711 Thanks for the help with this! The new functionality is published and available with `@golevelup/nestjs-rabbitmq@1.15.0`

perf2711:
Happy to see it merged! Will surely help my own project.

WonderPanda:
Glad to hear it! Please let me know if you come across any issues or area where it could be improved as you integrate it into your project

PR #135, opened 2020-04-17
feat(rabbitmq): add rabbitmq message handling

WonderPanda:
These changes allow you to use user-defined parser and serializer for messages passed through RabbitMQ.


WonderPanda:
Some things that still need to get updated before this can go through:
- finish integrating this with the new error handling logic that was recently introduced
- include some integration tests that verify the serialization/deserialization behavior
- cleanup the placement/naming of some interfaces and functions
- probably some docs

underfisk:
I'm closing/archiving this, it seems to be outdated and may not work as expected. Feel free to revive this if we'd like to get it across the finish line

PR #155, opened 2020-05-25
feat(hasura): add HasuraAction type

n3n:
Reference: https://hasura.io/docs/1.0/graphql/manual/actions/action-handlers.html#http-handler


WonderPanda:
Hey @n3n appreciate you opening this. I've been trying to brainstorm what the most idiomatic approach to Actions in NestJS would be. Maybe you'd like brainstorm with me in #156 about how to make Actions work in an intuitive way beyond the typings that you've added here

Issue #173, opened 2020-08-11
nestjs-rabbitmq: Is there any way to read headers & properties of message in consumer?


yogeshkathayat:
nestjs-rabbitmq: Is there any way to read headers & properties of message in pubsub consumer ?

WonderPanda:
Hi @yogeshkathayat, the raw message is passed to your consumer as a second argument. If you update the method signature of your consumer to accept `amqplib.ConsumeMessage` as a second argument you should be able to access all information that is available on the original message

rennenc:
Would be great to have this appear in the documentation.

PR #232, opened 2021-02-12
Improve documentation to explain how to deal with the original amqp message in message handlers

neilime:
It may help some developers

### DIFF
--- a/packages/rabbitmq/README.md
+++ b/packages/rabbitmq/README.md
@@ -227,7 +227,11 @@ The default behavior is that handlers will be attached, but to opt out simply se
 
 ### Dealing with the amqp original message
 
-In some scenarios, it wil be usefull the get the original amqp message (to retrieve the fields, properties...)
+In some scenarios, it wil be usefull the get the original amqp message (to retrieve the fields, properties...).
+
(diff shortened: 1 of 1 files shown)


packages/rabbitmq/README.md
```diff
@@ -225,6 +225,28 @@ In some scenarios, it may not be desirable for all running instances of a NestJS
...
+
+@Injectable()
+export class MessagingService {
+  @RabbitSubscribe({
+    exchange: 'exchange1',
+    routingKey: 'subscribe-route',
+    queue: 'subscribe-queue',
+  })
+  public async pubSubHandler(msg: {}, amqpMsg: ConsumeMessage) {
+    console.log(`Correlation id: ${amqpMsg.properties.correlationId}`);
```
underfisk, line 249: I think the 2nd argument you can add in the description that its going to be injected (optionally)
Would it really nice to add the reference (https://github.com/golevelup/nestjs/issues/173)

neilime: @underfisk I have updated to doc, does it fit with your feedback ?

underfisk (approved):
Overall is good and has no breaking changes

PR #252, opened 2021-05-03
docs: propose to use a function for deferred

psteinroe:
When using deferred as a static prop, problems will arise in setups where you reuse across multiple apps from a configurable core module, since externallyConfigured is called without the module actually being imported by another module.


WonderPanda:
@steinroe Thanks for catching this and sorry I missed getting this merged in. The function approach is definitely preferable here

WonderPanda (approved):

PR #260, opened 2021-05-11
fix(rabbitmq): prevent unhandled promise rejection warning

### LINKED ISSUE #259: UnhandledPromiseRejectionWarning when sending message on closed connection
If you use the `amqpConnection.request` when there is no connection yet you get a UnhandledPromiseRejectionWarning.

To reproduce config the RabbitMQModule with a broken url and wait = false
```
            exchanges: [
                {
                    name: 'amq.topic',
                    type: 'topic',
                },
            ],
            uri: 'amqp://invalid:invalid@localhost:5672',
        }),
```

Then use the `amqpConnection.request` like this:
```
try {
            const response = await this.amqpConnection.request<string>({
                exchange: 'amq.topic',
                routingKey: 'some_topic',
                payload: 'example',
                timeout: 10000, 
            });
        } catch (e) {
            console.log("error", e);           
        }
```

Now you will first get an UnhandledPromiseRejectionWarning before the catch logs the  "Error: Failed to receive response within timeout of 10000ms"

As far as i can see this is because the connection.ts `request` method is not try..catching the `publish` function.
The `publish` function throws an error when there is no connection

arjenvdhave:
Catch and rethrow errors from the publish function to prevent UnhandledPromiseRejectionWarning

fix #259

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -214,7 +214,7 @@ export class AmqpConnection {
     } catch (error) {
       //Catching errors from the publish function to prevent UnhandledPromiseRejectionWarning
       //Rethrow error so the caller of this function can catch the error.
-      throw e;
+      throw new Error(error.message);
     }
 
     const timeout$ = interval(timeout).pipe(
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -22,7 +22,6 @@ import {
   MessageHandlerErrorBehavior,
 } from './errorBehaviors';
 import { Nack, RpcResponse, SubscribeResponse } from './handlerResponses';
-import e = require('express');
 
 const DIRECT_REPLY_QUEUE = 'amq.rabbitmq.reply-to';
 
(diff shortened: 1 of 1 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -201,11 +202,21 @@ export class AmqpConnection {
...
+        payload,
+        {
+          replyTo: DIRECT_REPLY_QUEUE,
+          correlationId,
+        }
+      );
+    } catch (error) {
+      //Catching errors from the publish function to prevent UnhandledPromiseRejectionWarning
+      //Rethrow error so the caller of this function can catch the error.
+      throw e;
```
WonderPanda: @arjenvdhave This isn't the error, its an import for `express`.

WonderPanda (changes requested):

packages/rabbitmq/src/amqp/connection.ts
arjenvdhave: ah damn, sorry my vscode made this weird import

WonderPanda:
@arjenvdhave Looking at this some more, I don't think we actually need the try/catch here. All that is required for the error to bubble up properly is the `await` in front of `this.publish` so that the promise isn't treated in a fire and forget manner.

arjenvdhave:
Yes the try catch seems unnecessary indeed, will remove it

WonderPanda (approved):

PR #277, opened 2021-06-09
updated readme section Requesting Data from an RPC replaced '=' with :

nelsonBlack:
updated section Requesting Data from an RPC removed replaced '=' with :


WonderPanda:
@nelsonBlack Thanks!

PR #284, opened 2021-07-01
feat(hasura): metadata v3 support

WonderPanda:
## Improvements
Adds support for the new metadata v3 format

## Todo Before Merge
- [x] Add tests for v2 metadata to ensure we don't break things for existing users
- [x] Add tests for cron events


WonderPanda:
@BraunreutherA If you feel like having a look

BraunreutherA:
@WonderPanda thanks for your work on this. Very well implemented!

BraunreutherA:
Is there a reason why the test fixtures are in the `.gitignore`? I think then we would have to generate them on every test run.

WonderPanda:
> Is there a reason why the test fixtures are in the `.gitignore`? I think then we would have to generate them on every test run.

@BraunreutherA If they are not added to .gitignore then they will generate changes to the repo everytime someone runs the tests. I feel like it would be confusing for people pulling down the codebase if they have a fresh copy of the repo and then run tests and see a whole bunch of changes in git.

I've updated the logic to properly copy over the `.tmpl` files at test startup

Issue #316, opened 2021-09-24
Subscribe to queues without knowing the exchange and binding key (e.g. default exchange)


j-schreiber:
Hey all, I am having real troubles understanding how golevelup expects me to know the exchange and routing key in advance, when I simply want to consume messages from a particular queue. As I understood it, the  `@RabbitSubscribe` decorator requires an `exchange` and `routingKey`, as shown in the example above:

```javascript
@RabbitSubscribe({
    exchange: 'exchange1',
    routingKey: 'subscribe-route',
    queue: 'subscribe-queue',
  })
  public async pubSubHandler(msg: {}) {
    console.log(`Received message: ${JSON.stringify(msg)}`);
  }
```

I understand, that golevelup dynamically creates the binding. But what happens, if there is no exchange? If any publisher publishes to `amq.direct` (which is also the "empty string"), I have a valid queue, but neither an exchange nor binding (because rabbit routes the message implicitly and the queue name actually is the binding).

How am I supposed to subscribe to a typical pub-sub model where multiple exchanges and bindings might publish to the same queue? I have a landscape where the individual consumers are not supposed to know about other publishers and also do not care about the RabbitMQ configuration that routes messages to "their" queues. As long as they consume from the queues, they do not need to know about other exchanges and bindings.

Any help appreciated!

WonderPanda:
I think you probably want to use a `direct` exchange. https://www.rabbitmq.com/tutorials/tutorial-four-javascript.html

> The routing algorithm behind a direct exchange is simple - a message goes to the queues whose binding key exactly matches the routing key of the message.

I'm a bit confused about some of your other requirements though. It sounds like you're still required to know the names of queues ahead of time so there's still some burden of "knowing" the the destinations for things.

j-schreiber:
Hey @WonderPanda  thanks for responding so quickly! Yes, the direct exchange is useful, but in most of my use cases it's not necessary. Instead, I would like to use the [default exchange](https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchange-default) (which is either "empty string" or `amq.direct`). In that scenario, the publisher implicitly knows about the destination queue (even though it technically only specifies a binding key).

The queues are known ahead of time, yes. But to be precise, my question is: Can I subscribe to a queue without declaring an exchange and binding key? I simply want to consume everything from a particular queue, without worrying about the exchanges and bindings that route messages to that queue. Also, I do not want to filter. So I imagine something like this:

```javascript
[...]

PR #317, opened 2021-09-28
#316: Make exchange and routing key optional

j-schreiber:
I tried to prepare a PR for the topics discussed in #316 . I encountered an additional error when I was working with queues that already existed on the server, so I thought it would be a good idea to dynamically switch between `assertQueue` and `checkQueue`.

- Introduce additional option to assert or check queue
- Only bind queues, if exchange **and** routing key(s) is provided

Please help me improve this. I was not able to run the tests, they just failed with "cannot find module @golevelup/nestjs-common". Probably did something wrong setting up the sub-project.

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -412,16 +412,24 @@ export class AmqpConnection {
     subscriptionOptions: MessageHandlerOptions,
     channel: amqplib.ConfirmChannel
   ): Promise<string> {
-    const { exchange, routingKey } = subscriptionOptions;
-    let queue: unknown;
+    const {
+      exchange,
+      routingKey,
+      createQueueIfNotExists = true,
+    } = subscriptionOptions;
+    let actualQueue: unknown;
 
-    if (subscriptionOptions.createQueueWhenNotExisting) {
-      queue = await channel.assertQueue(
+    if (createQueueIfNotExists) {
+      const { queue } = await channel.assertQueue(
         subscriptionOptions.queue || '',
         subscriptionOptions.queueOptions || undefined
       );
+      actualQueue = queue;
     } else {
-      queue = await channel.checkQueue(subscriptionOptions.queue || '');
+      const { queue } = await channel.checkQueue(
+        subscriptionOptions.queue || ''
+      );
+      actualQueue = queue;
     }
 
     const routingKeys = Array.isArray(routingKey) ? routingKey : [routingKey];
@@ -430,12 +438,12 @@ export class AmqpConnection {
       await Promise.all(
         routingKeys.map((routingKey) => {
           if (routingKey) {
-            channel.bindQueue(queue as string, exchange, routingKey);
+            channel.bindQueue(actualQueue as string, exchange, routingKey);
           }
         })
       );
     }
 
-    return queue as string;
+    return actualQueue as string;
   }
 }
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -417,7 +417,8 @@ export class AmqpConnection {
       routingKey,
       createQueueIfNotExists = true,
     } = subscriptionOptions;
-    let actualQueue: unknown;
+
+    let actualQueue: string;
 
     if (createQueueIfNotExists) {
       const { queue } = await channel.assertQueue(
@@ -444,6 +445,6 @@ export class AmqpConnection {
       );
     }
 
-    return actualQueue as string;
+    return actualQueue;
   }
 }
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -408,12 +408,13 @@ export class AmqpConnection {
     return handler(message, msg);
   }
 
-  private async setupQueue<T>(
+  private async setupQueue(
     subscriptionOptions: MessageHandlerOptions,
     channel: amqplib.ConfirmChannel
-  ) {
+  ): Promise<string> {
     const { exchange, routingKey } = subscriptionOptions;
-    let queue;
+    let queue: unknown;
+
     if (subscriptionOptions.createQueueWhenNotExisting) {
       queue = await channel.assertQueue(
         subscriptionOptions.queue || '',
@@ -422,16 +423,19 @@ export class AmqpConnection {
     } else {
       queue = await channel.checkQueue(subscriptionOptions.queue || '');
     }
+
     const routingKeys = Array.isArray(routingKey) ? routingKey : [routingKey];
+
     if (exchange && routingKeys) {
       await Promise.all(
-        routingKeys.map((x) => {
-          if (x) {
-            channel.bindQueue(queue, exchange, x);
+        routingKeys.map((routingKey) => {
+          if (routingKey) {
+            channel.bindQueue(queue as string, exchange, routingKey);
           }
         })
       );
     }
-    return queue;
+
+    return queue as string;
   }
 }
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -254,18 +254,7 @@ export class AmqpConnection {
     msgOptions: MessageHandlerOptions,
     channel: amqplib.ConfirmChannel
   ): Promise<void> {
-    const { exchange, routingKey, allowNonJsonMessages } = msgOptions;
-
-    const { queue } = await channel.assertQueue(
-      msgOptions.queue || '',
-      msgOptions.queueOptions || undefined
-    );
-
-    const routingKeys = Array.isArray(routingKey) ? routingKey : [routingKey];
-
-    await Promise.all(
-      routingKeys.map((x) => channel.bindQueue(queue, exchange, x))
-    );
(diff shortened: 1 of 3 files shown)


j-schreiber:
I would be very happy for instructions how I can test my changes.

WonderPanda:
@j-schreiber I don't think we should be switching away from `assertQueue`. It will only throw an error if the queue already exists but with different options than what is being provided in the decorator. I think the solution is to make sure that the options actually match so that the source of truth is more easily managed in code.

In order to run the integration tests use the following steps:

Install dependencies: `yarn`
Build the projects: `yarn build`
Install e2e dependencies: `yarn --cwd ./integration/rabbitmq`
Ensure rabbitmq service is running: `docker-compose up -d`
Run the tests: `yarn test:integration`

j-schreiber:
Mh, I see. How could we handle the following situation: We use a dedicated repository to store the RabbitMQ configuration. The config has all queues, all exchanges and all bindings. We use RabbitMQ for the purpose of integrating enterprise applications, so it is quite "stable" (consumers do not dynamically register themselves and are not supposed to create queues).

Additionally, we recently made the move to quorum queues. Each queue has it's own dead-letter-routing key (but for now, they share the same dead letter exchange). We have multiple adapters (one for each proprietary system) that connect to Rabbit.

I prefer a single source of truth for my RabbitMQ above scattering it over 5+ repositories (one for each adapter). I want to be able to independently recreate my rabbit without the need to start adapters in order to configure it.

We are trying to use this library with one adapter (the other ones directly utilise amqlib and implement the subscription by themselves). This adapter now broke, because I introduced dead-letter exchange and quorum queues -- But it could easily kept working, if it just consumes from the queue without being so opinionated about it's configuration ;). 

So the question is: Is my approach considered a bad practice? If so, why?

WonderPanda:
Fair enough, you've convinced me. If you've got a centralized configuration for managing your Rabbit topology I can see how it would be annoying to have to ensure that the decorators match. Let's go with a property called `createQueueIfNotExists` instead

j-schreiber:
We found some old/deprecated dependencies in the project. Would you mind if we update them? If you do, please let me know so we can keep the change as minimal as possible.

Also, I am not able to build the package, using your instructions. We will investigate and check if I did something wrong during setup or if we might need to change something. Can you assist? See screenshot below:
image

E1: nvm ... Being the noob I am, I tried to run the command within the package, not the root.

WonderPanda:
> We found some old/deprecated dependencies in the project. Would you mind if we update them? If you do, please let me know so we can keep the change as minimal as possible.
> 
> E1: nvm ... Being the noob I am, I tried to run the command within the package, not the root.

No worries! I should have clarified that all commands are run from the root of the project.

I'd prefer to open multiple PRs to keep the changes smaller and easier to review if you'd like to also modify dependencies. 

October starts tomorrow so could also be a great way for you to easily get enough contributions to complete [Hacktoberfest](https://hacktoberfest.digitalocean.com/) and get some free swag!

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -429,4 +407,31 @@ export class AmqpConnection {
...
+    if (subscriptionOptions.createQueueWhenNotExisting) {
+      queue = await channel.assertQueue(
+        subscriptionOptions.queue || '',
+        subscriptionOptions.queueOptions || undefined
+      );
+    } else {
+      queue = await channel.checkQueue(subscriptionOptions.queue || '');
+    }
+    const routingKeys = Array.isArray(routingKey) ? routingKey : [routingKey];
+    if (exchange && routingKeys) {
```
WonderPanda, line 438: I think this condition might be wrong. You'll still need to bind the queue even if no specific exchange was provided but it should fall back to the default exchange. Otherwise, you won't be able to receive any messages

j-schreiber: Thx, I just got the tests running so I will check & fix that.

j-schreiber:
What would be the best approach: creating the queue (having the original behavior) if the new property is `undefined` or treating `undefined` as `false`?

I tend to the first option, so the update should be fully upwards compatible because it doesn't change the behavior. As long as you do not explicitly set it to `false`, it uses `assertQueue`. Or would it be better to change the option to something like `onlyCheckForQueue`? 

```javascript
    if (subscriptionOptions.createQueueIfNotExists || subscriptionOptions.createQueueIfNotExists === undefined) {
      queue = await channel.assertQueue(
        subscriptionOptions.queue || '',
        subscriptionOptions.queueOptions || undefined
      );
    }
```

looks a little awkward to me, tbh.

packages/rabbitmq/src/amqp/connection.ts
j-schreiber: All tests pass now. Logically, that makes sense to me. If the queue is already bound (because we invoked `assertQueue` earlier or because the binding already exists, we should be fine. I wasn't able to create tests for the use case, yet.

Can you hint me to where I need to change the decorator `RabbitSubscribe`? I cannot find the code where exchange and routing key are made required (in `MessageHandlerOptions` they are all optional).

WonderPanda:
> What would be the best approach: creating the queue (having the original behavior) if the new property is `undefined` or treating `undefined` as `false`?
> 
> I tend to the first option, so the update should be fully upwards compatible because it doesn't change the behavior. As long as you do not explicitly set it to `false`, it uses `assertQueue`. Or would it be better to change the option to something like `onlyCheckForQueue`?
> 
> ```js
>     if (subscriptionOptions.createQueueIfNotExists || subscriptionOptions.createQueueIfNotExists === undefined) {
>       queue = await channel.assertQueue(
>         subscriptionOptions.queue || '',
>         subscriptionOptions.queueOptions || undefined
>       );
>     }
> ```
> 
> looks a little awkward to me, tbh.

We should stick with the original behavior so that we don't introduce breaking changes to existing users of the package.

I would just destructure with a default value:

```typescript
const { createQueueIfNotExists = true } = subscriptionOptions;
```

packages/rabbitmq/src/amqp/connection.ts
WonderPanda: We will definitely need to add new e2e tests to verify that this all works as intended. Given the current code I really don't think you'll receive any messages since no binding is being created. You need to actually bind the queue, doing `check` or `assert` won't actually attach your handler.

Not sure what you mean about `RabbitSubscribe`, the changes to `MessageHandlerOptions` should be all that is required. Maybe you need to rebuild the project if you're getting Typescript errors? You can just use `yarn build:watch` if you want to have everything kept up to date automatically as you make changes

j-schreiber: When I use the `@RabbitSubscribe` decorator (as done in all the tests), `exchange`, `routingKey` and `queue` are required. I cannot find the code that does so. I believe it may be this snippet, but I am not good enough with TypeScript to fully understand what that code does (coming from a strongly typed proprietary language, so all this is pretty new for me)

```javascript
export const makeRabbitDecorator = <T extends Partial<RabbitHandlerConfig>>(
  input: T
) => (
  config: Pick<RabbitHandlerConfig, Exclude<keyof RabbitHandlerConfig, keyof T>>
) => (target, key, descriptor) =>
  SetMetadata(RABBIT_HANDLER, { ...input, ...config })(target, key, descriptor);
```

I will add new e2e tests to verify original behavior and to cover the new behavior. It just takes more time, because I am learning all this on the fly :D.

Thx for your patience and support, btw!

WonderPanda: @j-schreiber The code you've shown here is not really related. The only necessary change is to mark the properties in `MessageHandlerOptions` as optional:

```typescript
export interface MessageHandlerOptions {
  exchange?: string;
  routingKey?: string | string[];

 // rest of the interface
}
```

I just pulled down a fresh copy of the repository and made the change myself to make sure and it works as intended. All of the decorators no longer require that you specify `exchange` or `routingKey` as long as you make the change that I've shown.

You need to actually build the project in order for the changes to be picked up in the e2e test app though. Either run `yarn build` after you make a change or just use `yarn build:watch` to be able to have everything recompile on the fly while you make changes

j-schreiber:
@WonderPanda kindly review my changes, please. I added two e2e tests and I was able to verify, that the original behavior is unchanged (all tests green).

packages/rabbitmq/src/amqp/connection.ts
j-schreiber: I could have sworn that I did that several times. Now it worked.

WonderPanda:
> @WonderPanda kindly review my changes, please. I added two e2e tests and I was able to verify, that the original behavior is unchanged (all tests green).

Awesome! At a quick glance this change looks good. I'll try my best to do pull down your changes and do a proper review this afternoon. I appreciate you taking the time to dive in and make this change yourself

j-schreiber:
Hey @WonderPanda do you have any feedback on the status of my changes? Don't want to come off impatient; just want to know if I need to improve / add tests / etc?

WonderPanda:
> Hey @WonderPanda do you have any feedback on the status of my changes? Don't want to come off impatient; just want to know if I need to improve / add tests / etc?

This looks good. I made a couple quick changes that you can see in https://github.com/golevelup/nestjs/commit/2147352af020ceaa33f629b06096ec6264039bff

WonderPanda (approved):

Issue #315, opened 2021-09-23
[Question] Multiple handlers for a queue separated by routing key


vinaybedre:
Hey,

Is it possible to do something like following with golevelup/rabbitmq within the same module?
```typescript
@RabbitSubscribe({exchange:"any", queue:"user", routingKey:"key1"})
handleKey1() {
...
}

@RabbitSubscribe({exchange:"any", queue:"user", routingKey:"key2"})
handleKey2() {
...
}
```

vinaybedre:
If this is not possible, does it mean that by using golevelup/rabbitmq, we can only tie one method to an entire queue? 
What is the reasoning behind this? Should I need to use different queues for various purpose of communication between 2 parties?

WonderPanda:
Queues in RabbitMQ are lightweight and they are intended to be purpose driven. There is no benefit to to reusing the same queue for multiple handlers. A typical application that's properly leveraging RabbitMQ would have dozens if not hundreds or more queues.

A single RabbitMQ node could potentially handle tens of thousands of queues, see for example https://stackoverflow.com/a/22990200/1364771

If you *really* feel the need to do this (I'd be curious about your reasoning) you could have a single handler that receives all your messages through one queue. The handler could then do conditional processing based on the contents of the message itself.

There are no plans to try and implement the suggested functionality you've shown in your original comment as its generally an anti-pattern

vinaybedre:
@WonderPanda Thanks for the insights.
I have couple of microservices (8+), which communicate between them either via RPC / Event publishing.
I thought of having 1 queue for 1 service. 
I am afraid that having 1 queue per purpose may explode the number of queues I may have say in next 3 to 5 years. (just an estimation).

On the other hand, I do not want to have a single handler per service and use switch/if statements, where I will loose the context of typings. (For eg. I am using an interface/class which is used by producer and consumer to keep the input/outputs types strongly typed, at least in TypeScript) With 1 queue, 1 handler, I will be perfectly using strong types when communication happens between services.

WonderPanda:
I don't think number of queues should be a concern for you at all unless you think that it could increase to be hundreds of thousands or millions

eduscrakozabrus:
@RabbitSubscribe({
    exchange: 'messenger',
    routingKey: 'user-send-message',
    queue: 'messenger',
  })

и 
  @RabbitSubscribe({
    exchange: 'messenger',
    routingKey: 'customer-send-message',
    queue: 'messenger',
  })

[...]

PR #325, opened 2021-10-19
Fixed async callback

stgogm:
Callbacks should never be async as they can cause unexpected behavior due to crossed references.

### DIFF
--- a/packages/rabbitmq/src/rabbitmq.module.ts
+++ b/packages/rabbitmq/src/rabbitmq.module.ts
@@ -122,8 +122,9 @@ export class RabbitMQModule
 
     for (const key of providerKeys) {
       this.logger.log(`Registering rabbitmq handlers from ${key}`);
+
       await Promise.all(
-        grouped[key].map(async ({ discoveredMethod, meta: config }) => {
+        grouped[key].map(({ discoveredMethod, meta: config }) => {
           const handler = this.externalContextCreator.create(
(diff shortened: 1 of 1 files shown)


WonderPanda:
> Callbacks should never be async as they can cause unexpected behavior due to crossed references.

Which references might be crossed in this example? I don't currently see any issues with how its written

stgogm:
> > Callbacks should never be async as they can cause unexpected behavior due to crossed references.
> 
> Which references might be crossed in this example? I don't currently see any issues with how its written

You're right, it should work fine. Still, the `async` there is not really necessary.

On a side note, I started diggin in the code because I'm getting messages on the wrong handlers for some `routingKey`s. Might open another issue for that when I understand better why.

WonderPanda:
> On a side note, I started diggin in the code because I'm getting messages on the wrong handlers for some `routingKey`s. Might open another issue for that when I understand better why.

Are you re-using the same queue with multiple bindings like in https://github.com/golevelup/nestjs/issues/315? This is a pretty common mistake that I see people making

stgogm:
I see, then according to that, I attached multiple subscriptions to the same handler with the different routing keys (because there can be many that I don't care to handle if I use `#` or `prefix.*`) and it worked fine.

I don't know if this is the optimal pattern but it works.

Thanks for you quick and helpful answer! Cheers!

PR #329, opened 2021-10-28
feat(rabbitmq): allow amqplib 0.8.0 range

jeremylvln:
The new Node.js LTS is v16.
`amqplib` 0.7.x requires Node.js `<15`
`amqplib` 0.8.0 removes this constraint and requires `>=10`

This PR allows any project using the RabbitMQ package to use Node LTS again :)


WonderPanda:
Are there any breaking changes or concerns with us just updating it to 0.8.0 and dropping 0.7?

jeremylvln:
> Are there any breaking changes or concerns with us just updating it to 0.8.0 and dropping 0.7?

I kept the `^0.7.0` range to avoid breaking existing projects. As you can see [here](https://github.com/squaremo/amqp.node/releases/tag/v0.8.0), there is no functional change other than the Node version. If you want me to drop this range, I'll do!

WonderPanda:
I'd rather not support two versions at once. Let's just upgrade it to 0.8

jeremylvln:
> I'd rather not support two versions at once. Let's just upgrade it to 0.8

I've edited the patch, thanks!

WonderPanda:
Hmmm, the build failed looks like the image I was using might be deprecated in Azure Pipelines. Been meaning to switch over to GH Actions for a while but haven't found the time. I'd like to see that all the checks pass before I merge this

WonderPanda (approved):

WonderPanda:
@IamBlueSlime This is available in @golevelup/nestjs-rabbitmq@1.18.1

jeremylvln:
> @IamBlueSlime This is available in @golevelup/nestjs-rabbitmq@1.18.1

Awesome, work perfectly! Thank you

PR #333, opened 2021-11-03
One more typo in rabbitmq README

aaabramov:


aaabramov:
I am not sure why the build fails.

WonderPanda:
> I am not sure why the build fails.

The one that failed was just an optional check that attempts to attach a label to the PR

Issue #343, opened 2021-12-15
Setting of "arguments" in "queueOptions" object not working correctly


jlally21:
Hello,

I am setting `arguments` in the `queueOptions` object like:

```
  @RabbitSubscribe({
    exchange: "EVENT-BUS-MAIN-HEADERS-EXCHANGE",
    routingKey: '',
    queue: "TEST_QUEUE",
    queueOptions: {
      arguments: { 'x-match': 'all' },
    },
  })
  ```
  
 Both the queue ("TEST_QUEUE") and bind to exchange ("EVENT-BUS-MAIN-HEADERS-EXCHANGE") are created, but the arguments remain empty.
 `arguments: { 'x-match': 'all' }` Is that the correct format?  
 
 I tried finding example code online that sets the `arguments` field, but could not find any.  
 

![image](https://user-images.githubusercontent.com/17992893/146103187-3d5be4bd-ce15-4da9-93a7-ab44accdbf5e.png)

**Comment by kasir-barati** [comment]
I am getting this weird error message when I am trying to get my app send messages using headers exchange. Do you have any idea @jlally21?

```cmd
ERROR [AmqpConnection] Received message with invalid routing key
```

[Here is a repo for this issue](https://github.com/kasir-barati/nestjs-headers-exchange-rabbitmq).

PR #346, opened 2021-12-17
feat: add option for arguments when binding queue to an exchange

jlally21:
currently, there is no option to pass args to the channel.bindQueue function.  Passing arguments to
this function is required for some exchange types and queue-exchange setups

re #343


WonderPanda (approved):

WonderPanda:
Thanks for the contribution!

WonderPanda:
Available in `@golevelup/nestjs-rabbitmq@1.19.0`

Issue #348, opened 2021-12-22
changing queueOptions will crash next restart


christopheblin:
Let's say you have a subscriber like this

```
@RabbitSubscribe({
  exchange: 'dummy',
  queue: 'my_queue' 
})
```

You push the app to prod, everything is fine (exchange created if it does not exist, queue created if it does not exist, binding betwen queue and excahnge)

now, you change an option like deadLetterExchange

```
@RabbitSubscribe({
  exchange: 'dummy',
  queue: 'my_queue',
  queueOptions: {
    deadLetterExchange: 'dead-letter-retry'
  }
})
```

The app will not start :  
```
Operation failed: QueueDeclare; 406 (PRECONDITION-FAILED) with message "PRECONDITION_FAILED - inequivalent arg 'x-dead-letter-exchange' for queue 'my_queue' in vhost '/': received the value 'dead-letter-retry' of type 'longstr' but current is none"
```

I know this is the expected behavior from RabbitMq perspective (i.e multiple consumers may be bound to the queue so it is a bit unsafe to delete+recreate the queue automatically during assertQueue)

However, do you think this is the expected behavior with `@RabbitSubscribe` ?

I mean, if I explicitly specify the queue, it's because I want the same queue for all running replicas ... (besides, options do not really influence the running behavior of the queue from consumer point of view)

Will you accept a PR where we catch the failure of assertQueue to call deleteQueue + assertQueue ?

In any case, what do you suggest as a workaround to detect the situation automatically and to automatically delete the related queues before this lib calls assertQueue ?

WonderPanda:
These types of situations are tricky for a general purpose library because what might be appropriate for your scenario (in terms of delete + recreate) could be terrible for someone else.

I think that the best option (although it increases API surface) is to have an additional option inside `RabbitSubcribe` that allows you to specify custom behavior/error handling for when asserting the queue fails.

I'd definitely be open to accepting a PR that gives users more flexibility in this area.

christopheblin:
WDYT about this :

```
@RabbitSubscribe({
  exchange: 'dummy',
  queue: 'my_queue',
  queueOptions: {
    deadLetterExchange: 'dead-letter-retry'
  },
  assertQueueErrorHandler: (err, channel, queueName, queueOptions) => {
    if (/*err is precondition failed*/) {
       await channel.deleteQueue(queue)
       const { queue } = await channel.assertQueue(queue, queueOptions);
        return queue
    }
[...]

PR #349, opened 2021-12-22
fix(#348): assertQueueErrorHandler initial impl

christopheblin:
provide an error handler to handle failure during queue creation (see #348)


underfisk:
@christopheblin LGTM but there seems to be a change on the underlying amqlib package, do we really need that version update?

christopheblin:
@underfisk 

you're talkkng about the yarn.lock ? 

inside the file package.json (https://github.com/golevelup/nestjs/blob/master/packages/rabbitmq/package.json#L41), you already have (this is not me here)
```
"amqplib": "^0.8.0",
```

I jut ran yarn install and it automatically upgraded the yarn.lock according to what is specified by package.json ...

underfisk:
> 

Just wanted to be sure, long time i don't touch the package but I'll take a look at this PR soonish

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -416,16 +417,25 @@ export class AmqpConnection {
...
     let actualQueue: string;
 
     if (createQueueIfNotExists) {
-      const { queue } = await channel.assertQueue(
-        subscriptionOptions.queue || '',
-        subscriptionOptions.queueOptions || undefined
-      );
-      actualQueue = queue;
+      const queueName = subscriptionOptions.queue || '';
+      const queueOptions = subscriptionOptions.queueOptions || undefined;
```
underfisk: If `queueOptions` is not defined its going to be undefined so using `||` operator is not necessary

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -416,16 +417,25 @@ export class AmqpConnection {
...
 
     let actualQueue: string;
 
     if (createQueueIfNotExists) {
-      const { queue } = await channel.assertQueue(
-        subscriptionOptions.queue || '',
-        subscriptionOptions.queueOptions || undefined
-      );
-      actualQueue = queue;
+      const queueName = subscriptionOptions.queue || '';
```
underfisk: Use `??` operator instead of `||`

underfisk (changes requested):
LGTM, address the changes and I'll merge this

packages/rabbitmq/src/amqp/connection.ts
christopheblin: should be good

packages/rabbitmq/src/amqp/connection.ts
christopheblin: should be good

WonderPanda:
This has been published and is available in `@golevelup/nestjs-rabbitmq@1.20.0`

PR #350, opened 2021-12-22
use context type rmq instead of http

MetinSeylan:
if you use global nestjs exception handler or interceptor, rmq request context marked as an HTTP request but it's not, people try to handle rmq requests with HTTP methods and gets null pointer exception.


underfisk:
@MetinSeylan Thanks for submitting it, indeed your point is correct. I'll run some tests and then i think its good to be merged

underfisk:
Closed #359

PR #356, opened 2022-01-09
feat: add support for multiple channels

danocmx:
Adds support for multiple channels on 1 connection. Allows changing `prefetchCount` so handlers can process messages at different speeds. 

Should be backwards compatible so if application doesn't create more channels or handlers don't specify one, it will just use default one, which it creates. You can also optionally specify which channel is default and in that case it will be the only default channel in use.

Exchanges and `DirectReplyToQueue` are always set up on default channel.

Integration tests pass on my local machine and I have added one more to test this functionality.
## Usage
Module  registration:
```ts
import { RabbitMQModule } from '@golevelup/nestjs-rabbitmq';

@Module({
  imports: [
      exchanges: [
        {
          name: 'exchange1',
          type: 'topic',
        },
      ],
      uri: 'amqp://rabbitmq:rabbitmq@localhost:5672',
      channels: {
        'channel-1': {
          prefetchCount: 5,
        },
        'channel-2': {
          prefetchCount: 10,
          // Marks this channel as default for other handlers
          default: true,
        },
        'channel-3': {}
      },
      // This count is used when channel didn't specify one.
      prefetchCount: 25,
    }),
  ],
})
export class RabbitExampleModule {}
```

Message handlers:
```ts
import { RabbitSubscribe, RabbitRPC } from '@golevelup/nestjs-rabbitmq';
import { Injectable } from '@nestjs/common';

@Injectable()
export class MessagingService {
  @RabbitRPC({
    exchange: 'exchange1',
    routingKey: 'subscribe-route',
    queue: 'subscribe-queue',
    queueOptions: {
      channel: 'channel-1'
    }
  })
  public async rpcHandler(msg: {}) {
    console.log(`Received message: ${JSON.stringify(msg)}`);
    
    return { message: 'hi' };
  }
  
  // Uses default channel, because there isn't one specified.
  @RabbitSubscribe({
    exchange: 'exchange1',
    routingKey: 'subscribe-route-2',
    queue: 'subscribe-queue-2',
  })
  public async pubSubHandler(msg: {}) {
    console.log(`Received message: ${JSON.stringify(msg)}`);
  }
}
```

Sending message to subscriber:
```ts
// Both choose default channel to send the message over.

this.amqpConnection.publish('exchange1', 'subscribe-route-2', { message: 'hi' });

await this.amqpConnection.request({
  exchange: 'exchange1',
  routingKey: 'subscribe-route',
  payload: { message: 'ola' },
})
```


underfisk:
@danocmx Thanks for implementing this feature, I'll review this but at first glance it looks good to go

underfisk:
So after testing this locally and making sure this does not bring any breaking changes, it's a "go" from me but there is something I'm curious about first

I tried to remove the channels initialization from the e2e integration and see if they still work as expected to the correct channels. Turns out they go to the default channel that is initialized which I'm not sure whether is expected fallback behavior or not.

Having documentation update would be a starter and also some inline JSDoc for the new property explaining a little bit what it does

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -74,10 +78,24 @@ export interface RabbitMQConfig {
   connectionManagerOptions?: amqpConnectionManager.AmqpConnectionManagerOptions;
   registerHandlers?: boolean;
   enableDirectReplyTo?: boolean;
+  channels?: Record<string, RabbitMQChannelConfig>;
```
underfisk, line 98: Having a JSDoc would be beneficial
Updating the documentation to explain this new behaviour will also be a good thing

underfisk (approved):
So after testing this locally and making sure this does not bring any breaking changes, it's a "go" from me but there is something I'm curious about first

I tried to remove the channels initialization from the e2e integration and see if they still work as expected to the correct channels. Turns out they go to the default channel that is initialized which I'm not sure whether is expected fallback behavior or not.

packages/rabbitmq/src/rabbitmq.interfaces.ts
danocmx: If there aren't any channels specified, it always spawns a default one. This is done for backwards compability so applications don't break if they don't create their own channels.

I'll try to add some documentation when I have time.

underfisk: @danocmx Yes i get it that we need a default one but if we publish to channel 2 and we don't have a channel 2, it seems to be routing to the default, is that the expected behaviour?

danocmx: Yes, that is expected behaviour. I made it publish from default channel because exchanges are asserted on it.

underfisk: Thanks for confirming, once the documentation is addressed we can merge

packages/rabbitmq/README.md
```diff
@@ -95,6 +96,8 @@ Import and add `RabbitMQModule` it to the `imports` array of module for which yo
 
 If you are using exchanges, provide information about them to the module and they will be automatically asserted for you as part of initialization. If you don't, it's possible message passing will fail if an exchange is addressed that hasn't been created yet.
 
+You can also optionally create your own channels which you consume messages from. If you don't create your own channels there will always be one created by default. You can also select which channel is default if you are creating your own. By setting `prefetchCount` for a particular channel you can manage message speeds of your various handlers on the same connection.
```
underfisk, line 99: Could we also mention the `default` property inside a channel definition, that would probably help when people are looking to define the default custom channel

WonderPanda:
Thanks for the contributions guys! This is available in `@golevelup/nestjs-rabbitmq@1.20.0`

PR #359, opened 2022-01-11
feat: RMQ context type + nestjs upgrade to v7.5.5 

underfisk:
Initially, there was request #350 that proposed to fix the `contextType` to RMQ instead of relying on the default `http` avoiding null pointer exceptions.
In fact that is true, when attempting to get anything from RMQ inside (lets say a guard) it will not return anything valid

Tested this change and it does not introduce any breaking changes and the Nestjs upgrade to v7 was required in order to have the newest features such as the `contextType`

Changelog:
- Silent logger for stripe webook e2e tests
- Context type is now using `rmq` instead of the default `http`
- Upgraded nestjs packages to v7.5
- Dev dependencies upgrades to its latest such as Jest
- Fixed error with hasura event handling decorators

**Note**
There are some temporary eslint disables that i hope after upgrading the majority of the packages to the most safe and recent version that we can tune them


underfisk:
Closing in favor of #342

PR #371, opened 2022-01-21
Migrate to Nx monorepo

### LINKED ISSUE #143: Convert to using NX Monorepo

jeremylvln:
This PR converts this Lerna monorepo to a Nx one. Closes #143.
**Note: this PR is based on #342.**

I've done no functional changes except lint fixes. I work with Nx every day, so I guess this implementation is somewhat ideal. Sorry for the large diff, a lot of files has been moved.

I've modified the CI accordingly, do not worry about the `MansaGroup/nrwl-nx-action` action used inside, I've created it :)

Tests are welcomed!


underfisk:
@IamBlueSlime Thanks a bunch for doing this, our priority is getting v8 PR first and then this should be our next step

jeremylvln:
> @IamBlueSlime Thanks a bunch for doing this, our priority is getting v8 PR first and then this should be our next step

That's why this PR is rebased on the other one :D

WonderPanda:
Wondering if it would be easier to just do a fresh migration over to NX now that the upgrade to Nest 8 is done. There's a huge number of duplicated commits and merge commits here that we don't really need.

jeremylvln:
I will rebase the PR correctly asap

jeremylvln:
Hi @WonderPanda, @underfisk. I've rebased the PR, I hope that I've not removed any commit during the process. As this PR moves a lot of file, pay attention to the modification of these file if I've missed something.

Btw, I've removed the CI job name, thus the required checks will not be filled, this PR will need to be force merged by an administrator.

packages/common/project.json
```diff
@@ -28,6 +28,19 @@
         "jestConfig": "packages/common/jest.config.js",
         "passWithNoTests": true
       }
+    },
+    "version": {
+      "executor": "@jscutlery/semver:version",
```
WonderPanda, line 33: @IamBlueSlime Do you have experience working with these plugin before as part of your day job? Or are these being included based just on the discussion we were having in #143

packages/rabbitmq/src/lib/amqp/connection.ts
```diff
@@ -259,17 +259,12 @@ export class AmqpConnection {
...
-      requestOptions.routingKey,
-      payload,
-      {
-        replyTo: DIRECT_REPLY_QUEUE,
-        correlationId,
-        headers: requestOptions.headers,
-        expiration: requestOptions.expiration,
-      }
-    );
+    this.publish(requestOptions.exchange, requestOptions.routingKey, payload, {
```
WonderPanda, line 262: Are you skipping the git hooks locally?  It seems like this is reverting Prettier's formatting.

Also, why is the `await` being dropped here and on line 389 below?

jeremylvln: I'm not skipping the hooks, it seems that this is the "new way to go". I've added a `lint` CI job which is passing, so I don't think there is an issue here.

If an `await` was removed, it's surely because the called function was not async, otherwise, it would have thrown a lint error in either case :)

packages/common/project.json
jeremylvln: Nope, I don't have any on these plugins, I just saw your discussion in the issue while doing this PR, so I've added them. But by reading their README's it seemed pretty accurate to our needs here :/

underfisk: @IamBlueSlime Could you potentially give us some pro/cons of using this rather than other available plugins or ways you would achieve this? I haven used this plugin before but I'm kinda curious

WonderPanda (changes requested):
I'm not interested right now in switching away from Yarn 1 and over to NPM so we should revert the package-lock.json and changes to the CI scripts.

In the future I'd be more open to considering PNPM but it should be as a different PR and not mixed in with the NX stuff

packages/common/project.json
WonderPanda: Okay no worries, I'll want to do some tests to make sure that the release and changelog process is smooth before we merge this in but the plugins definitely look promising

jeremylvln: @underfisk I never used Nx to publish packages, I'm afraid I can't give you any pro/cons about any package concerning publishing :/

underfisk:
> I'm not interested right now in switching away from Yarn 1 and over to NPM so we should revert the package-lock.json and changes to the CI scripts.
> 
> In the future I'd be more open to considering PNPM but it should be as a different PR and not mixed in with the NX stuff

I agree, if we have to switch our package manager I would definitely advocate for **PNPM** but now on this scope, we should just aim to switch to Nx with the bare minimum of what we have, and then we can start thinking about what's next

underfisk:
@IamBlueSlime Can we merge the latest changes, i think we should start considering having Nx asap, there are some stuff that we have pending that can actually leverage having this ecosystem

jeremylvln:
> @IamBlueSlime Can we merge the latest changes, i think we should start considering having Nx asap, there are some stuff that we have pending that can actually leverage having this ecosystem

Hi @underfisk! There is currently an issue with Nx that breaks JS reflection on build, see https://github.com/nrwl/nx/pull/8784. I do not think that libraries are impacted, but only the apps that are build. In all cases, just to be sure, I think it would be preferable to wait until my PR on Nx is merged before merging this one.

But if you want me to rebase this PR and remove npm-related commits, I can totally do that :)

underfisk:
> > @IamBlueSlime Can we merge the latest changes, i think we should start considering having Nx asap, there are some stuff that we have pending that can actually leverage having this ecosystem
> 
> Hi @underfisk! There is currently an issue with Nx that breaks JS reflection on build, see [nrwl/nx#8784](https://github.com/nrwl/nx/pull/8784). I do not think that libraries are impacted, but only the apps that are build. In all cases, just to be sure, I think it would be preferable to wait until my PR on Nx is merged before merging this one.
> 
> But if you want me to rebase this PR and remove npm-related commits, I can totally do that :)

No rush, i've ran into a similar problem before, well at max you can override terser plugin with a custom nx webpack file but if they can include that officially it would be better

underfisk:
@IamBlueSlime I've noticed that the Nx PR has been merged, any news so far if it is included in any of the latest versions?

underfisk:
@IamBlueSlime Do you have any plans on having this? I believe we were close to merge this experience

jeremylvln:
> @IamBlueSlime Do you have any plans on having this? I believe we were close to merge this experience

Hey, sorry for the long time, I've been quite busy at work. I do not think I will have the time to work on this today, but maybe tomorrow or Wednesday. Is this okay for you?

Issue #379, opened 2022-01-24
RabbitMQ logging - Unable to replace RabbitMQ logger with an external logger


shakedo:
In RabbitMQ module, ConsoleLogger (@nestjs/common) is used for logging and by that it's not possible to replace it with an external logger as described here: https://docs.nestjs.com/techniques/logger#use-external-logger

The solution is to use Logger (@nestjs/common) instead of ConsoleLogger in these files:

packages\rabbitmq\src\amqp\connection.ts
packages\rabbitmq\src\rabbitmq.module.ts

underfisk:
@shakedo You can extend `ConsoleLogger` and we shouldn't be going back to the deprecated `Logger`, you can see at: https://docs.nestjs.com/techniques/logger#extend-built-in-logger that you can extend and then provide your logger

shakedo:
Thois code for example, dosent replace RabbitMQ logs and do replace NestJs logs:

class MyLogger extends ConsoleLogger {
  log(message: any, context?: string) {
    super.log(message + '!!!!!!!', context);
  }
}

async function bootstrap() {
  const app = await NestFactory.create(AppModule, { bufferLogs: true });
  app.useLogger(new MyLogger());
  app.flushLogs();
  await app.listen(3000);
}
bootstrap();



[Nest] 12380  - 24.01.2022, 16:50:28     LOG [NestFactory] Starting Nest application...!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [NestFactory] Starting Nest application...!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [InstanceLoader] LoggerModule dependencies initialized!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [NestFactory] Starting Nest application...!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [NestFactory] Starting Nest application...!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [InstanceLoader] LoggerModule dependencies initialized!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [InstanceLoader] DiscoveryModule dependencies initialized!!!!!!!
[Nest] 12380  - 24.01.2022, 16:50:28     LOG [InstanceLoader] AppModule dependencies initialized!!!!!!!
[...]

PR #380, opened 2022-01-24
Replace ConsoleLogger with Logger so it can be replaced by an external logger

shakedo:
This PR is for solving this issue: https://github.com/golevelup/nestjs/issues/379

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -1,4 +1,4 @@
-import { ConsoleLogger } from '@nestjs/common';
+import { Logger } from '@nestjs/common';
 import {
   ChannelWrapper,
   AmqpConnectionManager,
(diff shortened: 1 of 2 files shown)


WonderPanda (approved):
This looks fine to me. @IamBlueSlime @underfisk do you know if there was a particular reason we switched to using ConsoleLogger instead as part of the upgrade to Nest 8?

jeremylvln:
No idea about this change, I didn't do it personally.

underfisk:
@WonderPanda Since v8 `Logger` is now deprecated, `ConsoleLogger` is the new standard logger, I don't think we should revert because the reason behind the issue was that there was no way of customizing the logger using `ConsoleLogger` but there is, extending the logger is not tied to the any specific class such as `Logger`

underfisk (changes requested):
This is a regression and there's no need for us to use deprecated services

WonderPanda:
Indeed it seems like this is the way forward as of Nest 8: https://docs.nestjs.com/migration-guide#logger-breaking-changes

PR #388, opened 2022-01-31
Revert logger usage

### LINKED ISSUE #379: RabbitMQ logging - Unable to replace RabbitMQ logger with an external logger

underfisk:
@WonderPanda After confirming with Jay it seems that `ConsoleLogger` should only be used at a consumer level which means, as the producers need to extend the `Logger` service. The docs do not explain this for module creators but this is intended and after confirming it I don't believe it is a regression

This will fix #379

### DIFF
--- a/packages/hasura/src/hasura.module.ts
+++ b/packages/hasura/src/hasura.module.ts
@@ -2,7 +2,7 @@ import { DiscoveryModule, DiscoveryService } from '@golevelup/nestjs-discovery';
 import { createConfigurableDynamicRootModule } from '@golevelup/nestjs-modules';
 import {
   BadRequestException,
-  ConsoleLogger,
+  Logger,
   Module,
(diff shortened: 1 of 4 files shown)


WonderPanda:
There are a number of other places where the code was switched to use ConsoleLogger. Do you mind also updating those as well?

underfisk:
> There are a number of other places where the code was switched to use ConsoleLogger. Do you mind also updating those as well?

Will do

PR #389, opened 2022-01-31
feat: RabbitMQ module drop support for build method

underfisk:
BREAKING CHANGE: Any app that still uses build should use forRoot or forRootAsync

### DIFF
--- a/packages/rabbitmq/src/rabbitmq.module.ts
+++ b/packages/rabbitmq/src/rabbitmq.module.ts
@@ -62,25 +62,6 @@ export class RabbitMQModule
     return connection;
   }
 
-  public static build(config: RabbitMQConfig): DynamicModule {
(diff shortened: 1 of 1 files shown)


WonderPanda:
I'm not sure if I want to release another breaking change version so shortly after the last one. It doesn't hurt to leave `build` in for a little bit longer while we figure out next steps in case there are any other breaking changes coming down the pipeline. I'd rather we release all the breaking changes together as version 3

underfisk:
> I'm not sure if I want to release another breaking change version so shortly after the last one. It doesn't hurt to leave `build` in for a little bit longer while we figure out next steps in case there are any other breaking changes coming down the pipeline. I'd rather we release all the breaking changes together as version 3

If we aim to drop this at v3 I'm fine with that, its just that having consistency across nestjs community packages leveraging `forRoot` and `forRootAsync` only would be ideal

Issue #369, opened 2022-01-21
RabbitMQ controllers support


underfisk:
Previously I had suggested that we migrate to microservices adapter, well the way that microservices packages work with what we have as services is using controllers which gives the developer more flexibility to leverage things such as pipes, guards, etc etc

My proposal here is to support both services and controllers and help people to slowly migrate (having two ways of defining handlers etc can be confusing but i want to make it as simple as possible)

If anyone appreciates the idea, has any feedback at all just let me know, i haven't yet started but this doesn't take that much to do

jeremylvln:
It would be neat indeed, one step forward hexagonal architecture \o\

verti4cal:
+1

otherwise i have to rewrite any service function or have to implement a second function as a wrapper.

```
  async userFindOneByUuidQuery(query: QueryUserFindOneByUuid): Promise<User> { // used for the rpc
    return this.userModel.findOne({ uuid: query.uuid });
  }

  async userFindOneByUuid(uuid: string): Promise<User> { // used internal
    return this.userModel.findOne({ uuid: uuid });
  }
```

PR #390, opened 2022-02-01
feat(rabbitmq): implemented controller discovery

verti4cal:
re #369

implemented option to enable discovery of handlers in controllers.

### DIFF
--- a/integration/rabbitmq/e2e/rpc-controller-discovery.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/rpc-controller-discovery.e2e-spec.ts
@@ -1,0 +1,93 @@
+import { AmqpConnection } from '@golevelup/nestjs-rabbitmq';
+import { INestApplication } from '@nestjs/common';
+import { Test } from '@nestjs/testing';
+import { AppModule } from '../src/app.module';
+
+const nonJsonRpcRoutingKey = 'non-json-rpc-2';
+
+describe('Rabbit Controller RPC', () => {
+  let app: INestApplication;
+  let amqpConnection: AmqpConnection;
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  afterAll(async () => {
+    await app?.close();
+  });
+
+  beforeAll(async () => {
+    const moduleFixture = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    amqpConnection = app.get<AmqpConnection>(AmqpConnection);
+    await app.init();
+  });
+
+  it('regular RPC handler should receive a valid RPC response', async () => {
+    const response = await amqpConnection.request({
+      exchange: 'exchange2',
+      routingKey: 'rpc-2',
+      payload: {
+        request: 'val',
+      },
+    });
+
+    expect(response).toEqual({ echo: { request: 'val' } });
+  });
+
+  it('intercepted RPC handler should receive a transformed RPC response', async () => {
+    const response = await amqpConnection.request({
+      exchange: 'exchange2',
+      routingKey: 'intercepted-rpc-2',
+    });
+
+    expect(response).toEqual({ transformed: { message: 42 } });
+  });
+
+  it('error reply RPC handler with non-JSON message should return an RPC error response', async () => {
+    const response = await amqpConnection.request({
+      exchange: 'exchange2',
+      routingKey: 'error-reply-rpc-2',
+      payload: Buffer.from('{a:'),
+    });
+
+    expect(response).toHaveProperty('message');
+    expect(response).toMatchObject({ status: 'error' });
+  });
+
+  it('non-JSON RPC handler with should receive a valid RPC response', async () => {
+    const response = await amqpConnection.request({
+      exchange: 'exchange2',
+      routingKey: nonJsonRpcRoutingKey,
+      payload: {
+        request: 'val',
+      },
+    });
+
+    expect(response).toEqual({ echo: { request: 'val' } });
+  });
+
+  it('non-JSON RPC handler with undefined message should receive a valid RPC response', async () => {
+    const response = await amqpConnection.request({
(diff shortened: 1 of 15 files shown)


underfisk:
@verti4cal The problem with this approach will be how the events are registered, for example if you use a guard or a pipe, does it actually run on nest lifecycle? We want to leverage controllers because of the value that they provide
I'm working on actually migrating to the microservices by building a custom transport for us, the problem now is that a PR is officially under review which will provide us the metadata we just need but if you can make sure interceptors, pipes, guards work (writing e2e tests will be a good way to assert that) then i'm fine considering this possibility

verti4cal:
Ok, i added an e2e test to check the interceptors.
It works, but only if the interceptor is an instance not a class.

I could not find the cause of this in the nestjs-source.

underfisk:
> feat(rabbitmq): added e2e-tests for interceptors on controller-discovery

I suspect because if you pass the class then you have to instantiate and do that using the context creator but this tests just cover interceptor, could you also add an example using Guard and Pipes?

verti4cal:
I found the problem with the instance and class error on the Interceptor.
The controller has to be defined as provider in the module.

I also added tests for Guards.

But i cant get Pipes to run.
They wont get called.
Porbably they get only called in HTTP-Requests, because according to nestjs-Docs they only work with HTTP-Decorators (Param, Body, etc.).

underfisk:
> I found the problem with the instance and class error on the Interceptor. The controller has to be defined as provider in the module.
> 
> I also added tests for Guards.
> 
> But i cant get Pipes to run. They wont get called. Porbably they get only called in HTTP-Requests, because according to nestjs-Docs they only work with HTTP-Decorators (Param, Body, etc.).

If you take a look at the source code, in microservices package you can see that they use a resolver to allow pipes to be invoked, they are not tied to HTTP only, as I said adding controllers support isn't so straightforward due to the context fact. 
I appreciate the initiative but if we are about to allow providers to be used then we need to make sure the whole lifecycle comes along

verti4cal:
> If you take a look at the source code, in microservices package you can see that they use a resolver to allow pipes to be invoked, they are not tied to HTTP only, as I said adding controllers support isn't so straightforward due to the context fact. I appreciate the initiative but if we are about to allow providers to be used then we need to make sure the whole lifecycle comes along

Thanks for the hint with the microservices source.
I implemented it like they did in nestjs-microservices.

underfisk:
> feat(rabbitmq): implemented pipes for controller-discovery

I mean there's progress in your last PR, it aligns with what i was saying about making sure the context creator needs to run custom pipes, guards etc. 
Before i step into this more deeply, can you research a little bit more on passing guards/interceptors as classes so that we can resolve them normally? The reason i'm asking is because in guards, sometimes we need services injected (lets say my guard wants to validate a JWT or something) and if we enforce to create an instance, then the DI won't work

verti4cal:
I have already found the problem.
Sorry, I only briefly pointed it out in the old comment.

The problem was, I had to add the controller to the providers in the module.

```
@Module({
  imports: [
      useFactory: () => ({
        exchanges: [
          {
            name: 'exchange2',
            type: 'topic',
          },
        ],
        uri,
        connectionInitOptions: { wait: true, reject: true, timeout: 3000 },
      }),
    }),
  ],
  controllers: [ControllerDiscoveryController],
  providers: [ControllerDiscoveryController],
})
export class ControllerDiscoveryModule {}
```

As soon as I did this, I could use the classes of Guards, Pipes, Interceptors.

underfisk:
> I have already found the problem. Sorry, I only briefly pointed it out in the old comment.
> 
> The problem was, I had to add the controller to the providers in the module.
> 
> ```
> @Module({
>   imports: [
>     RabbitMQModule.forRootAsync(RabbitMQModule, {
>       useFactory: () => ({
>         exchanges: [
>           {
>             name: 'exchange2',
>             type: 'topic',
>           },
>         ],
>         uri,
>         connectionInitOptions: { wait: true, reject: true, timeout: 3000 },
>         enableControllerDiscovery: true,
>       }),
>     }),
>   ],
>   controllers: [ControllerDiscoveryController],
>   providers: [ControllerDiscoveryController],
> })
> export class ControllerDiscoveryModule {}
> ```
> 
> As soon as I did this, I could use the classes of Guards, Pipes, Interceptors.

This is great!
What happens if you have a global module with the RabbitMQModule + the controllers in another module OR if RabbitMQModule has imported once in an app.module but you have a user.module (with user.controller) + another module (with another controller). I want to be sure that we don't need to enforce loading the controllers within the same module definition, if that works then i'll be glad to consider this in our next release

verti4cal:
> This is great! What happens if you have a global module with the RabbitMQModule + the controllers in another module OR if RabbitMQModule has imported once in an app.module but you have a user.module (with user.controller) + another module (with another controller). I want to be sure that we don't need to enforce loading the controllers within the same module definition, if that works then i'll be glad to consider this in our next release


Everything works fine.
I've added new tests for this case.

underfisk:
> > This is great! What happens if you have a global module with the RabbitMQModule + the controllers in another module OR if RabbitMQModule has imported once in an app.module but you have a user.module (with user.controller) + another module (with another controller). I want to be sure that we don't need to enforce loading the controllers within the same module definition, if that works then i'll be glad to consider this in our next release
> 
> Everything works fine. I've added new tests for this case.

When i have time I will evaluate, just wanted to thank you for doing this PR

WonderPanda:
@verti4cal This looks really comprehensive! Thanks for a great contribution

@underfisk Did anything jump out to you in this one that warrants further investigation or do you think I should be good to merge now?

underfisk:
@WonderPanda It's under a flag so it won't interfere with any existing code that's providers, I still need a run a few tests but it looks good to me.
If this is much more accepted than providers, i will advocate to remove the flag and then turn this as the default option later

underfisk:
@WonderPanda Ship it, LGTM!

underfisk (approved):

PR #405, opened 2022-03-01
fix(rabbitmq): update uuid version

HHongSeungWoo:
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use
Math.random() in certain circumstances, which is known to be problematic.  See
https://v8.dev/blog/math-random for details.


jeremylvln:
Maybe we could use the built-in UUID generator provided by Node 16. Now that it is the new LTS it should not be an issue and one dependency less

https://nodejs.org/api/crypto.html#cryptorandomuuidoptions

HHongSeungWoo:
If this repo only supports version 14.17, 16 or higher, that's good idea, I will fix it.
But I don't know support version range.

underfisk:
@WonderPanda Do we want to drop support on versions behind LTS (in this particular case v16)?

WonderPanda:
> @WonderPanda Do we want to drop support on versions behind LTS (in this particular case v16)?

I think its reasonable for us to only support LTS, so I'm fine to switch to using the built in crypto APIs for this

HHongSeungWoo:
@WonderPanda please review it

jeremylvln (approved):

underfisk (approved):
LGTM

jeremylvln:
Would be nice to remove Node 12 and 14 from the CI matrix (in this PR or another one ofc) :)

underfisk:
> Would be nice to remove Node 12 and 14 from the CI matrix (in this PR or another one ofc) :)

Yes, I think we should aim to drop support for those versions, nowadays I believe people uses at least Nodejs 14+, I would be down to at least remove Nodejs 12 so this PR would also make more sense to me

underfisk:
@WonderPanda This LGTM to be merged, I've made sure we start with Nodejs 14 as this following PR will contain #426, the CI 12 will fail because the crypto module we're using was introduced at `14.7`

underfisk:
@WonderPanda As soon as you can, try to merge this PR and a few others blocked by the staled CI (12)

HHongSeungWoo:
@WonderPanda I fixed the mistake, please try again

PR #411, opened 2022-03-30
feat(rabbitmq): add support for multiple connections

### LINKED ISSUE #393: How to open multiple connection and use in their respective modules?
Hello, is it possible to create multiple connection and use in their respective modules? I tried like

`@Module({
  imports: [
      exchanges: [
        {
          name: 'exchange1',
          type: 'headers',
        },
      ],
      uri: 'amqp://username:password@server1:5672',
    })
  ],
  controllers: [FirstController]
})
export class FirstController {}`

`@Module({
  imports: [
      exchanges: [
        {
          name: 'exchange2',
          type: 'headers',
        },
      ],
      uri: 'amqp://username:password@server2:5672',
    }),
  ],
  controllers: [SecondController],
})
export class SecondController {}`

it created 2 connection but same queues on both rmq server without respecting modules.

andrewda:
Fixes https://github.com/golevelup/nestjs/issues/393

WIP and still mostly untested, but wanted to put up a PR in case anyone has suggestions on how this can be improved. At a high-level, this adds a `name` property for each connection, and allows us to reference the specific connection in decorators. For example:

```ts
@Module({
  imports: [
      useFactory: () => ({
        name: 'connection1',
        exchanges: [
          {
            name: 'exchange1',
            type: 'topic',
          },
        ],
        uri,
      }),
    }),
  ],
  controllers: [AppController],
  providers: [SubscribeService],
})
export class AppModule {}
```

```ts
@Injectable()
class SubscribeService {
  @RabbitSubscribe({
    connection: 'connection1',
    exchange,
    routingKey: '#',
    queue,
  })
  handleSubscribe(message: object) {
    console.log(`RECEIVED MESSAGE: ${message}`);
  }
}
```

This also adds a new `AmqpConnectionManager` class and provider to access the array of connections. This is implemented similarly to [`nestjs-redis`](https://www.npmjs.com/package/nestjs-redis)'s method for accessing multiple connections: `this.amqpConnectionManager.getConnection('connection-name')`.

I'm also striving to keep this backwards-compatible, so as not to require any major bumps. All the existing unit and integration tests pass, so I think I've succeeded there, but I'll need to do some additional testing to make sure.


andrewda:
@underfisk the most recent commit should have cleaned this up a decent amount. When you have a few minutes to review, let me know what you think of this approach and whether you have any changes/suggestions!

packages/rabbitmq/src/rabbitmq.module.ts
```diff
@@ -104,86 +120,99 @@ export class RabbitMQModule
   }
 
   async onApplicationShutdown() {
-    this.logger.verbose('Closing AMQP Connection');
-    await this.amqpConnection.managedConnection.close();
+    this.logger.verbose('Closing AMQP Connections');
+
+    for (const connection of this.connectionManager.getConnections()) {
```
underfisk: Use promise all instead so u can close all in "parallel"

underfisk (changes requested):

packages/rabbitmq/src/amqp/connectionManager.ts
```diff
@@ -0,0 +1,19 @@
+import { AmqpConnection } from './connection';
+
+export class AmqpConnectionManager {
```
underfisk, line 3: One of the things we need to handle is disconnections, so lets say we have connection 1 and 2. If connection 2 disconnects we need to retry/attempt but when it just drops or becomes unreachable we need to remove from the active connections array otherwise its going to be considered as an active connection. What do you think? 🤔

packages/rabbitmq/src/rabbitmq.module.ts
```diff
@@ -122,9 +122,11 @@ export class RabbitMQModule
   async onApplicationShutdown() {
     this.logger.verbose('Closing AMQP Connections');
 
-    for (const connection of this.connectionManager.getConnections()) {
-      await connection.managedConnection.close();
-    }
+    await Promise.all(
+      this.connectionManager
+        .getConnections()
+        .map((connection) => connection.managedConnection.close())
```
underfisk: `.map((connection) => connection.managedConnection.close)` will be better because just passing the reference will make sure that the promise handler will invoke and await but Its just a suggestion

underfisk (changes requested):
LGTM to me overall, thanks for submitting this, I haven't yet ran the tests but we'll need them for sure to assert that everything still workings fine
Let's also wait for @WonderPanda review on this PR

packages/rabbitmq/src/amqp/connectionManager.ts
andrewda: Good questions. I would probably lean towards keeping the dropped connection in the connections array so the user is able to attempt manual reconnection with `amqp-connection-manager`. Since the only time a connection is added to the array is in the `AmqpConnectionFactory`, removing any of the connections will essentially remove it until the entire application is restarted.

underfisk: @andrewda What i would recommend then is to have some kind of state alongside with the connection such as `connected: boolean` to indicate whether the connection is alive or not (the name is just an example)

andrewda: Sure, I like that! We could add something to `AmqpConnection` like:

```ts
get connected() {
	this._managedConnection.isConnected();
}
```

Does that sound reasonable?

underfisk: Yes i like the idea

andrewda: Sweet, added in 30e5123fcbf299f2e80aa18837498eecdfa37ecd!

allhaker:
Would really like this to get merged =)
Would help us a lot!

WonderPanda:
Great work, I like the approach taken here. However it seems like the integration tests are failing in CI for some reason

andrewda:
Will take a look at those integration tests today!

andrewda:
@WonderPanda 34cca4ba547796e0d08e6bab184c6740d6c1cec2 should fix the failing integration tests, looks like there were 2 tests using the same exchange and queue names.

verti4cal:
Great work!

Updated today, but got a error on my project.
The AmqpConnection wont get injected anymore.
The AmqpConnectionManager has the running "default" connection.

Is someone experiencing the same issue ?

underfisk:
@verti4cal Can you tell us which version you were in before and you are now and also what versions are your nestjs dependencies, I'd like to know when it stopped working. Also if you can give a small repro, that would be good but we should avoid adding more here so please open a new github issue and we track the problem there

miguelcabgil:
Hey people. I hope you're doing well. Could anyone explain how could I specify the connection I need to connect to for publishing and the connection I want to use to subscribe in the controllers with the option enableControllerDiscovery as true? Thank you!

PR #416, opened 2022-04-07
feat(connection): reconsume

rphansen91:
When dependencies of a Microservice are offline it is beneficial to stop the consumer and only
restart when work is ready to be continued. 

This saves a lot of rabbit bandwidth not needing to deliver nack requeue over and over until the consumer is ready.

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -38,7 +38,6 @@ import { isNull } from 'lodash';
 
 const DIRECT_REPLY_QUEUE = 'amq.rabbitmq.reply-to';
 
-
 export type ConsumerTag = string;
 
 export type SubscriberHandler<T = unknown> = (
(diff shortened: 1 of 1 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -77,6 +100,9 @@ export class AmqpConnection {
   private _channel!: Channel;
   private _channels: Record<string, Channel> = {};
   private _connection?: Connection;
+
+  private _consumers: ConsumerHandler<unknown, unknown>[] = [];
```
WonderPanda: Can we switch using a `Map<string, ConsumerHandler>` instead and use the `consumerTag` as the key? It would simplify a lot of the other code so that there isn't a need to do things like `findIndex` and `splice`

WonderPanda (changes requested):

underfisk (changes requested):
@rphansen91 The only request i have here would be having tests to assert this feature
I'll merge this but I would love to actually have some coverage even though this does not break anything

PR #432, opened 2022-04-29
refactor: allow any type to be used with request function

kuskoman:
Right now the interface passed to request method needs to extend Record<string, unknown>, which is in many cases incorrect, since this function may return any JSON parseable object, for example array:

```json
'["a","b"]'
```

or undefined (yes, not really useful response, but a correct and parseable one)
```json
undefined
```

Removing requirement to extend Record<string, unknown> would allow user to return valid type without workarounds like this:
```ts
const rpcResponse: T | T[] = await this.amqpConnection.request<T>({ [...] })
if (Array.isArray(rpcResponse)) {
  return rpcResponse
}

throw new Error(...)
```

I also removed await, since both my human sight and static analisys tool stated that the function cannot return promise


WonderPanda (approved):

kuskoman:
> First-time contributors need a maintainer to approve running workflows.

@WonderPanda

underfisk:
> > First-time contributors need a maintainer to approve running workflows.
> 
> 
> 
> @WonderPanda 

I have approved the workflows to run

PR #440, opened 2022-05-25
fix:(hasura) Fixes schedule event handlers payload

ivorpad:
1. The schedule events payload doesn't have a `name` property, so the only way to identify these events is with `payload.comment`.

```js
{
  payload: {
    scheduled_time: "2022-05-25T15:13:00Z",
    payload: "{}",
    created_at: "2022-05-25T15:12:11.937494Z",
    id: "a1314019-ea6d-4461-988f-424418f6c2b4",
    comment: "",
  },
  headers: [
    {
      value: "application/json",
      name: "Content-Type",
    },
    {
      value: "hasura-graphql-engine/v2.6.2",
      name: "User-Agent",
    },
    {
      value: "myadminsecretkey",
      name: "nestjs-event-webhook",
    },
  ],
  version: "1",
}

```

https://hasura.io/docs/latest/graphql/core/api-reference/metadata-api/scheduled-triggers/#metadata-create-scheduled-event

So, this function will always return `false`

https://github.com/golevelup/nestjs/blob/bcb6fc6b4294ac2200f7b85a0c4ccd62bfec5782/packages/hasura/src/hasura.module.ts#L39-L43

In this fix we expect a `comment` to identify the event handler:

  ```ts
@HasuraEventHandler({
    triggerName: 'event_comment'
  })
  eventHandler() {
    console.log('eventHandler', new Date())
  }
```

```js
{
  payload: {
    scheduled_time: "2022-05-25T15:13:00Z",
    payload: "{}",
    created_at: "2022-05-25T15:12:11.937494Z",
    id: "a1314019-ea6d-4461-988f-424418f6c2b4",
    comment: "event_comment",
  },
  headers: [
    {
      value: "application/json",
      name: "Content-Type",
    },
    {
      value: "hasura-graphql-engine/v2.6.2",
      name: "User-Agent",
    },
    {
      value: "myadminsecretkey",
      name: "nestjs-event-webhook",
    },
  ],
  version: "1",
}
```

2. I also assigned `evt.id` to `keys` instead of `null`, because I think it's better to have some sort of key to debug.


packages/hasura/src/hasura.module.ts
```diff
@@ -182,6 +182,8 @@ export class HasuraModule
       if (handlers && handlers.length) {
         return Promise.all(handlers.map((x) => x.handler(evt)));
       } else {
+        console.log('key for error message', keys, typeof keys);
```
WonderPanda: Can you remove this or replace it with a a call through the actual logger if its needed

ivorpad (commented):
_<deleted>_

packages/hasura/src/hasura.module.ts
ivorpad: @WonderPanda This was a mistake. Removed.

PR #443, opened 2022-05-27
feat: serializer/deserializer

HHongSeungWoo:
Added for message compression and many other operations. I think it will be a useful feature.

*I don't speak English well. if don't understand, please reply


WonderPanda (approved):
This is a nice addition, thank you for your contribution!

WonderPanda:
This should be published and available in `@golevelup/nestjs-rabbitmq@2.4.0`

PR #457, opened 2022-06-17
feat(modules): display compiler error on missing attributes at forRootAsync

### LINKED ISSUE #456: Display compiler error on missing attributes at `forRootAsync`
Currently the compiler does not enforce you to provide values at `forRootAsync`.
It is valid just to call `Module.forRootAsync(Module, {})` even, if there is a needed config.

In my opinion the compiler should complain about that and therefore reduce possible runtime-errors.

_Note:_ I will provide a PR to fix that.

nidomiro:
feat(modules): display compiler error on missing attributes at forRootAsync

BREAKING CHANGE: it is required to provide one of 'useExisting', 'useClass' or 'useFactory' to forRootAsync

closes #456

### DIFF
--- a/packages/modules/src/dynamicModules.ts
+++ b/packages/modules/src/dynamicModules.ts
@@ -1,6 +1,5 @@
 import { DynamicModule, Provider, Type } from '@nestjs/common';
 import { ModuleMetadata } from '@nestjs/common/interfaces';
-import { get } from 'lodash';
 import { interval, lastValueFrom, race, Subject } from 'rxjs';
 import { first, map } from 'rxjs/operators';
 import { OptionalFactoryDependency } from '@nestjs/common/interfaces/modules/optional-factory-dependency.interface';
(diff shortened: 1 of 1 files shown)


WonderPanda:
I think in this case it could technically be considered a breaking change since it is modifying the public API surface but if anyone actually was not passing options their application would not have worked in the first place so I can't imagine there are people out there relying on the current (pre-fix) functionality.

For that reason I'm going to release this as a patch fix, otherwise we'll have to do a major version bump on all of the @golevelup packages that build on top of this one

nidomiro:
Do you need any action from me?

nidomiro:
I changed the commit to include the removal of the lodash import

nidomiro:
@WonderPanda when will this PR be merged? Or is there still an open point I have to change?

WonderPanda (approved):

PR #459, opened 2022-06-20
chore(rabbitmq): update amqp-connection-manager to v4

vincent-benbria:
This brings compatibility (maybe?) with **amqp-connection-manager** > 4.0.0

amqp-connection-manager [4.0.0](https://github.com/jwalton/node-amqp-connection-manager/compare/v3.9.0...v4.0.0) (2022-01-07)

**BREAKING CHANGES**
We will no longer emit a disconnect event on an
initial connection failure - instead we now emit **connectFailed** on each
connection failure, and only emit disconnect when we transition from
connected to disconnected.


underfisk:
@vincent-benbria Is there a way to proxy the old disconnect event, I'm trying to think of a way to bring compatibility to pre-existing code. If not, if we decide that this is a breaking change, please ensure we're only going to have 4.xx 

I did play with this for a while, this seems good to me but i don't want to either support v3/v4 when there's no proxy or whatsoever so bridge both v3 and v4.

underfisk:
@vincent-benbria Could you just make sure we're on `amqp-connection-manager` v4 and update the `yarn.lock` file?

andrewda:
@underfisk would it make sense to bump `amqplib` to `^0.10.1` in this PR as well? That should keep it in sync with the latest version of Nest:

https://github.com/nestjs/nest/blob/81df0dde0881b0680808798565bc88a0e65dc2ed/package.json#L108

If you'd rather that be in a separate PR, I can do that as well.

underfisk:
It does make sense, bump it

underfisk:
@vincent-benbria I have bumped `amqlib` as @andrewda suggested and made sure that we are only aiming for v4 from now forward, after reading this major it feels that its important to guide people through v4 using the new api methods

witem:
Any ETA for this to be merged?

PR #461, opened 2022-06-23
Set specific type names for our execution contexts

### LINKED ISSUE #454: [Stripe] Execution context in webhook handlers incorrectly has 'http' type
Firstly thanks for all your efforts with these modules, they're very helpful!

I have a global `APP_GUARD` configured in my NestJs application, which tries to fetch authorization information from headers on the HTTP request object. The guard was working perfectly until I added `@golevelup/nestjs-stripe` to my application.

The guard is now being triggered by Stripe webhook handlers. When this happens the NestJs execution context passed to the guard has type `http`, but the arguments array contains the Stripe webhook payload instead of the HTTP request and response that it should have. So when I do `context.switchToHttp().getRequest().headers.authorization` an error occurs because `headers` is undefined -- `getRequest()` is actually returning an object like `{ object: 'event', id: 'evt_XXX', api_version: '2020-08-27', etc }`.

image

I could easily work around this if the context of webhook handlers had a different type (e.g. `stripe_webhook`), but because it claims to be an HTTP context the only option is to resort to inspecting the shape of the context arguments, which feels kind of dirty.

The NestJs `ExternalContextCreator.create()` method accepts a ninth(!) argument which is the context type:
https://github.com/nestjs/nest/blob/cbec376216831303d581bf23c85fa7ad0d08f43e/packages/core/helpers/external-context-creator.ts#L91-L108

In your call this argument is omitted, so it receives the default value of 'http'. I think if you simply adjusted this call to specify a custom context type it would make life better for apps that use NestJs guards, filters and interceptors. [The relevant code is here.](https://github.com/golevelup/nestjs/blob/351b6f0882b0b910786ea702f332be0e4c4135cd/packages/stripe/src/stripe.module.ts#L130-L134)

I also note that the `@golevelup/nestjs-rabbitmq` and `@golevelup/nestjs-hasura` modules have the same issue, which is related to #204. The solution that was adopted there looks like it wouldn't have been necessary if a different context type was set instead.

sgarner:
Fixes #454

The NestJS `ExternalContextCreator.create()` method accepts a ninth(!) argument which sets a name for the context type. When omitted, it receives the default value `http`. This can make it difficult for NestJS enhancers (guards, interceptors, filters) to behave correctly when the contexts don't actually have the properties of a real HTTP context.

This PR changes the type for these contexts:

- Stripe module contexts now use the type `stripe_webhook`
- Hasura module contexts now use the type `hasura_event`

RabbitMQ module contexts were already using the type `rmq` since release 2.0.0 (4092c22316023ceacb4218cb30710cc14a4539bd).

### DIFF
--- a/packages/hasura/README.md
+++ b/packages/hasura/README.md
@@ -22,6 +22,7 @@ Leverage NestJS to make incorporating business logic and event processing easier
       - [Registering Scheduled Event Handlers](#registering-scheduled-event-handlers)
       - [Retry Configuration](#retry-configuration)
       - [Configuring Hasura Environment Variables](#configuring-hasura-environment-variables)
+      - [Usage with Interceptors, Guards and Filters](#usage-with-interceptors-guards-and-filters)
     - [Related Hasura Documentation](#related-hasura-documentation)
       - [Concepts](#concepts)
       - [Tutorials](#tutorials)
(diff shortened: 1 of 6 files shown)


WonderPanda (approved):

WonderPanda:
Thanks for this! I just need to figure out a release strategy since this should probably be considered a breaking change for the hasura and stripe packages

Issue #415, opened 2022-04-07
Cancel Consumer and Restart


rphansen91:
**Issue**
Currently it is possible to cancel a consumer, but their is not clear way to restart the consumer.

**Reason**
When dependencies of a Microservice are offline it is beneficial to stop the consumer and only restart when work is ready to be continued. This saves a lot of rabbit bandwidth not needing to deliver nack requeue over and over until the consumer is ready.

**Solution**
Maintain a list of the consumer options so that the canceled consumer can be restarted

craigotis:
@underfisk This feature seems to meet one of our service needs - but I'm finding it hard to retrieve the `ConsumerTag` created once a subscriber is established. The `AmqpConnection.createSubscriber()` method returns `Promise<void>`, so it's unclear what we're supposed to pass to the cancel/resume functions to make use of this. 🤔

underfisk:
@craigotis Would you be down to have a PR to fix this issue?

craigotis:
@underfisk I'm happy to look into it - though it seems the subscription is created in the `addSetup()` callback for the `ChannelWrapper` from [this project](https://github.com/jwalton/node-amqp-connection-manager/blob/master/src/ChannelWrapper.ts), so it would potentially need to be implemented as an optional callback parameter in the `createSubscriber()` function.

Before I go that route though, am I missing something? How was this cancel/restart implementation intended to be used? 

**Edit:** It might look something like this: https://github.com/golevelup/nestjs/pull/472

underfisk:
@craigotis Your PR looks good to me but i would advocate adding some test coverage so that we can safely track whether its properly working

PR #472, opened 2022-07-21
Returns the ConsumerTag from the createSubscriber() and createRpc() functions

craigotis:
This will enable the `ConsumerTag` to be subsequently passed to the `cancelConsumer()` and `resumeConsumer()` methods on the `AmqpConnection` class.

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -343,7 +343,7 @@ export class AmqpConnection {
         msgOptions,
         channel,
         originalHandlerName
-      )
+      );
       consumerTagCallback?.(consumerTag);
     });
   }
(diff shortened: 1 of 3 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -480,6 +486,8 @@ export class AmqpConnection {
       rpcOptions,
       channel,
     });
+
+    return consumerTag;
```
underfisk, line 507: My general recommendation is to return an object so that in the future if we decide to export more we can easily just add a new property

craigotis: @underfisk Good call, updated [here](https://github.com/golevelup/nestjs/pull/472/commits/1597daa6bd3de78b89eb5b98821eaeaf084d300c).

Are you able to approve the pending workflow, to give the added e2e tests a shot? (I'm having some difficulty spinning them up locally, they seem to be wired primarily for CI use?)

underfisk: @craigotis CI approved :)

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -405,20 +414,28 @@ export class AmqpConnection {
       msgOptions,
       channel,
     });
+
+    return consumerTag;
```
underfisk, line 419: What i meant is that the return type is changed to an object that has a `consumerTag` property, you can create a new interface for the return type so that it become easily scalable

craigotis: Right - I exposed it as an interface on any `public` methods when we're returning the tag to consumers, but left it as a single return value on any internal functions. (Since there's little risk to breaking/tweaking them in the future if needed.) Would you prefer it be wrapped in an interface for the `private` functions as well?

underfisk: Yes i saw that and it looks good for the callback but i also believe it would look okay in the return type of the function itself, if the previous type were `void` then we are not going to break anything since the current version does not expose anything. My point of view here is merely scalability, if people start consuming this right away and in the future, we want to explore more properties out of the channel then it would break right away every consumer

craigotis: Are you suggesting that the return type of `createSubscriber()` be updated from `Promise<void>` to `Promise<SubscriptionResult>`, or that the internal `setupSubscriberChannel()` return type be updated from `Promise<ConsumerTag>` to `Promise<SubscriptionResult>`? 

(Appreciate the incredibly prompt and helpful feedback, by the way!)

underfisk: I believe only the public API because the internal methods will not use it (but. yeah something like SubscriptionResult)

craigotis: Got it. The issue is we don't have the ConsumerTag at that point - our public method body looks like:
```
return this.selectManagedChannel(
      msgOptions?.queueOptions?.channel
    ).addSetup(async (channel) => {
      const consumerTag = await this.setupSubscriberChannel<T>(...);
    });
)
```
And the `consumerTag` comes _back to us_ inside a callback - not as a return value. Bit of context [here](https://github.com/golevelup/nestjs/issues/415#issuecomment-1191579258), but if there's a workaround or something I'm missing, I'd definitely also prefer to return `Promise<SubscriptionResult>` instead of `Promise<void>` with a callback.

**Edit:** We could do something like this, but that means that the `createSubscriber()` function would now wait for the channel to come up, instead of returning more quickly like it does today:
```
  public async createSubscriber<T>(
    handler: SubscriberHandler<T>,
    originalHandlerName: string,
    return new Promise((res) => {
      this.selectManagedChannel(
        msgOptions?.queueOptions?.channel
      ).addSetup(async (channel) => {
        const consumerTag = await this.setupSubscriberChannel<T>(
          handler,
          msgOptions,
          channel,
          originalHandlerName
        );
        res({ consumerTag })
      });
    });
  }
```

underfisk: I see. I'll leave this up to your judgment, I think people will probably come up with more ideas when consuming this so

craigotis: @underfisk I'm trying the `return Promise(...)` route, would you mind kicking off the CI? 🙏

craigotis:
@underfisk I got the integration tests running locally, so was able to verify a bit more before kicking up more dust.

I think the changes and tests are now ready for another re-review and CI run, if you're able. Thanks!

underfisk:
@craigotis Thanks for doing this, I'll merge this since it doesn't break anything and we can include this in the next version

craigotis:
@underfisk Wonderful, happy to help. Is there a schedule I can communicate to our team as to the project's release cadence?

underfisk:
@craigotis Only @WonderPanda would be able to publish a new release, I'm not sure of the actual plans to a next release yet

WonderPanda:
Hey guys I can publish a new release. Nice work! It was cool to see you guys put together this feature collaboratively today. This is a welcome addition to the library.

One thing I'll ask is just that we review the final commit message together because based on how things are currently set up all of the changelogs depend on us properly using [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/).

The right commit title here would probably have been something like `feat(rabbitmq): expose consumer tag for cancel and resume`

I'm going to rewrite the commit message on master and then do a release

WonderPanda:
@craigotis This is published and available now as `@golevelup/nestjs-rabbitmq@3.1.0`

I also update the [all contributors](https://github.com/golevelup/nestjs/commit/13a8fd9d4a381c5eed2dc623b2639eda51771e54) list so you get credit on the readme

Issue #447, opened 2022-06-08
Possibility to add custom error handler for timeouts


L-U-C-K-Y:
Hi @WonderPanda 

Thanks for all your work on these packages!

I am getting a lot of `Error: Failed to receive response within timeout of 10000ms` within one of our most heaviest microservices and I'm trying to find out which RPC calls are causing the timeouts. We have 500+ queues with each RPC having a queue and I would like to avoid increasing the timeout for all of them.

Do you see a possibility to add a custom error handler for timeouts, so that we could for example add the queue name, or some general context to debug this?

Thanks!

kzeratal:
Make the error message more clear should address the issue.

abz-denys-z:
Give me the opportunity to catch the error myself
I want to specify the instance of the class by which I will handle the error myself

PR #478, opened 2022-07-25
chore: more clear error message on reponse timeout

kzeratal:
According to this issue #447 .


abz-denys-z:
Give me the opportunity to catch the error myself
I want to specify the instance of the class by which I will handle the error myself

Issue #480, opened 2022-07-27
Custom logger is not used when using nest js 9


shakedo:
I'm using nest js 9.0.0 and when I'm using custom logger, rabbitmq client logs are written to default logger and not to my custom logger:

[Nest] 26531  - 07/27/2022, 9:36:17 AM     LOG [AmqpConnection] Trying to connect to RabbitMQ broker (default)
[Nest] 26531  - 07/27/2022, 9:36:17 AM     LOG [RabbitMQModule] Successfully connected to RabbitMQ
[Nest] 26531  - 07/27/2022, 9:36:17 AM     LOG [RabbitMQModule] Initializing RabbitMQ Handlers
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"NestFactory","msg":"Starting Nest application..."}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"InstanceLoader","msg":"DiscoveryModule dependencies initialized"}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"InstanceLoader","msg":"AppModule dependencies initialized"}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"InstanceLoader","msg":"LoggerModule dependencies initialized"}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"InstanceLoader","msg":"RabbitMQModule dependencies initialized"}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"RoutesResolver","msg":"AppController {/}:"}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"RouterExplorer","msg":"Mapped {/, GET} route"}
{"level":30,"time":1658903777435,"pid":26531,"hostname":"host","context":"NestApplication","msg":"Nest application successfully started"}
[Nest] 26531  - 07/27/2022, 9:36:17 AM     LOG [AmqpConnection] Successfully connected to RabbitMQ broker (default)
[Nest] 26531  - 07/27/2022, 9:36:17 AM     LOG [AmqpConnection] Successfully connected a RabbitMQ channel "AmqpConnection"

In main js, default logger is replaced:
  const app = await NestFactory.create(AppModule, { bufferLogs: true });
  app.useLogger(app.get(Logger));
  await app.listen(3000);

Issue is reproduced here: https://github.com/shakedo/nest9-example

alko89:
[...]

PR #481, opened 2022-07-27
chore: update to nest 9

alko89:


alko89:
Not sure how to test before merge, but it should fix #480

alko89:
Should fix #480

PR #483, opened 2022-07-27
fix(rabbitmq): properly close the AMQP connections on application shutdown

### LINKED ISSUE #482: Managed connections are not being closed on application shutdown
I believe an issue was introduced [here](https://github.com/golevelup/nestjs/pull/411/files#diff-a41b46ff3698a6e6a638ed448cdfabb193c623aae07af7bda9ed225432a5a2bbR129), and the `close()` method is no longer being called on the managed connections on shutdown.

craigotis:
Rather than returning a Promise from our map() callback, we were returning the function to be
called.

fix #482


WonderPanda (approved):

WonderPanda:
Thanks! Good catch

PR #490, opened 2022-08-04
fix(rabbitmq): fix usage of handler config

### LINKED ISSUE #489: RabbitMQ: "handlers" feature not working
The "handlers" feature introduced in https://github.com/golevelup/nestjs/pull/446 does not work correctly because the handler config is not actually used to create the subscription. Will provide a PR shortly.

chromey:
fix #489


underfisk:
@chromey Thanks for addressing this, was this something covered in any unit/e2e tests? If not, should we add some coverage to avoid regression?

chromey:
I gave it a shot. Took me some time to get the existing e2e cases to succeed consistently. I assume there are some race conditions happening. With `yarn test:integration --run-in-band` results are consistent

PR #495, opened 2022-08-11
fix(rabbitmq): handle connnection not available issue on RabbitSubscribe and RabbitRPC

vijay94:
handle connnection not available issue on RabbitSubscribe and RabbitRPC

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -343,6 +343,7 @@ export class AmqpConnection {
     originalHandlerName: string
   ): Promise<SubscriptionResult> {
     return new Promise((res) => {
+      let result: SubscriptionResult;
       this.selectManagedChannel(msgOptions?.queueOptions?.channel)
         .addSetup(async (channel) => {
           const consumerTag = await this.setupSubscriberChannel<T>(
@@ -351,9 +352,9 @@ export class AmqpConnection {
             channel,
             originalHandlerName
           );
-          res({ consumerTag });
+          result = { consumerTag };
         })
-        .then((result: any) => {
+        .then(() => {
           res(result);
         });
     });
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -343,8 +343,8 @@ export class AmqpConnection {
     originalHandlerName: string
   ): Promise<SubscriptionResult> {
     return new Promise((res) => {
-      this.selectManagedChannel(msgOptions?.queueOptions?.channel).addSetup(
-        async (channel) => {
+      this.selectManagedChannel(msgOptions?.queueOptions?.channel)
+        .addSetup(async (channel) => {
           const consumerTag = await this.setupSubscriberChannel<T>(
             handler,
             msgOptions,
@@ -352,8 +352,10 @@ export class AmqpConnection {
             originalHandlerName
           );
           res({ consumerTag });
-        }
-      );
+        })
+        .then((result: any) => {
+          res(result);
+        });
     });
   }
 


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -343,17 +343,19 @@ export class AmqpConnection {
...
             msgOptions,
             channel,
             originalHandlerName
           );
           res({ consumerTag });
-        }
-      );
+        })
+        .then((result: any) => {
+          res(result);
```
WonderPanda, line 358: I'm confused as to how we can resolve two values from the same promise. Doesn't the promise always get resolved before it can get to this new `then` continuation when we resolve the consumer tag on line 354?

WonderPanda (changes requested):

packages/rabbitmq/src/amqp/connection.ts
vijay94: Hi @WonderPanda , the addSetup function actually either calls the callback or return undefined. We have handled the call back but didn't handle the returned value.

https://github.com/jwalton/node-amqp-connection-manager/blob/a1092b3a3a8e866080a1b400cd826a8edb739e3d/src/ChannelWrapper.ts#L228

This is the code snippet of the addSetup function. Please let me know if the returned value can be handled differently.

vijay94:
@WonderPanda Any update on this.

davidtemplin:
I applied this patch locally using `patch-package` and it alleviated the issue.

packages/rabbitmq/src/amqp/connection.ts
davidtemplin: This patch worked when I applied it locally using `patch-package`. However, I wonder if there is a way to rewrite it such that there is only one invocation of `res`, e.g.:

```typescript
    originalHandlerName: string
    return new Promise((res) => {
      let result;
      this.selectManagedChannel(msgOptions?.queueOptions?.channel)
        .addSetup(async (channel) => {
          const consumerTag = await this.setupSubscriberChannel<T>(
            handler,
            msgOptions,
            channel,
            originalHandlerName
          );
          result = { consumerTag };
        }
      )
        .then(() => {
          res(result);
```

I'm not very familiar with the `promise-breaker` package and the `addSetup` function, so I am not sure if this is appropriate. Each `Promise` is a `Promise<void>`, hence the `result` variable.

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -343,17 +343,19 @@ export class AmqpConnection {
...
-        async (channel) => {
+      this.selectManagedChannel(msgOptions?.queueOptions?.channel)
+        .addSetup(async (channel) => {
           const consumerTag = await this.setupSubscriberChannel<T>(
             handler,
             msgOptions,
             channel,
             originalHandlerName
           );
           res({ consumerTag });
```
davidtemplin: This worked when I applied it locally. Consider only resolving the `Promise` in one place. See one suggestion in the [comment](https://github.com/golevelup/nestjs/pull/495/files#r966322388) below.

vijay94: Hi @davidtemplin the addSetup functions receives a callback as a parameter and returns a promise. If the connection is successful it calls the callback function we passed. And if the connection fails it returns some value. So I think we would have to resolve or reject in both the callback function and then block.

davidtemplin: Thank you, @vijay94. Perhaps I have misunderstood. The `addSetup` function is called without specifying any `done` parameter. Then, `addSetup` returns the result of calling `pb.addCallback` without any `done` parameter. According to the documentation of `pb.addCallback`, this simply returns the supplied `Promise`, namely, this `Promise` (let's call it "Promise A"):

```typescript
(this._settingUp || Promise.resolve()).then(() => {
                this._setups.push(setup);
                if (this._channel) {
                    return pb.call(setup, this, this._channel);
                } else {
                    return undefined;
                }
            })
```

Promise A proceeds as follows: first, Promise A waits for the `Promise` `(this._settingUp || Promise.resolve())` to resolve, next the `then` handler is called, and finally Promise A resolves.

Now, when the `then` handler is called, the `setup` callback will be called if and only if `this._channel` is defined. However, _in all cases_, Promise A resolves to `undefined` (at least at the type level, i.e. assuming that the `setup` callback really is a `void` callback). If the `then` handler invokes `return pb.call`, it will wait for the `Promise` returned from `pb.call` to resolve to a value, and then resolve to the same value (which should be `undefined` since the `setup` callback is supposed to be a `void` function). If the `then` handler immediately returns `undefined`, then `Promise A` resolves immediately - but regardless, it resolves to `undefined`; the only difference is whether or not the `setup` callback is invoked. Either way, when it is done, it resolves to `undefined`.

Thus, the patch that I suggested should work. When Promise A is finished, the final `then` handler is called. If the `setup` callback was invoked, a value will be set in the `result` variable. Otherwise, `result` will be `undefined`.

We can then resolve with `res(result)`. I wonder if an `error` handler is needed to reject the `Promise` or if errors cannot occur for some reason.

vijay94: Yeah @davidtemplin, I went through the documentations and your solution will work. Please feel free to push the code directly to the master of the forked repo.

davidtemplin: Thanks, @vijay94 ! I'll do that soon.

davidtemplin: @vijay94 - branch protection is enabled for the `master` branch of the forked repo. Would you like me to create a PR with a new branch or would you prefer to disable the branch protection?

vijay94: Hi @davidtemplin, Can you raise a PR. I will approve and merge it right away. Thanks

vijay94:
@WonderPanda Can you check the PR now?

davidtemplin:
@vijay94 - are you going to merge the `master` branch of your forked repository back into this branch or open a new PR with the changes we implemented?

@WonderPanda - will you be able to address this soon? I am using a patched version of the package meanwhile.

vijay94:
@davidtemplin Merged our changes to the fix branch. Lets wait for @WonderPanda to merge it.

WonderPanda (approved):

WonderPanda:
Hey this is now published and available in `@golevelup/nestjs-rabbitmq@3.3.0`

PR #503, opened 2022-09-11
fix(package.json): add stripe v10 as optional peerDependency 

### LINKED ISSUE #502: Update stripe package version
Currently, a peer dependency for stripe package is set to version 8. There is version 10 already that handles new API version. It would be good to add this version as an optional peerDependency.

ofilipowicz:
fix #502

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
     "@nestjs/platform-express": "^9.0.5",
     "lodash": "^4.17.21",
     "reflect-metadata": "^0.1.13",
-    "rxjs": "^7.5.6",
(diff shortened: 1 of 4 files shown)


underfisk:
I would like for us to only use the latest from now forward

ofilipowicz:
@underfisk Fine by me. I pushed a change with only the recent stripe version as ^10.8.0 in peerDependencies

underfisk:
@ofilipowicz You'll have to install the stripe as a dev dependency otherwise it can't build, check the CI build errors

PR #506, opened 2022-09-23
feat(rabbitmq): add the option to avoid declaring exchanges

frankmangone:
It is entirely possible for a rabbitmq user to not have rights to declare an exchange, and just consume it instead. This is the motivation of this PR: I couldn't find any option in the current state of the package that allows me to do this (only patching!).

Implementation is fairly simple: if I set the exchange creation flag to `false`, I should only check for its existence.

This is crucial for the project I'm working on, so I'd really appreciate if this is added as a feature. If it's not, I can always continue to patch, but this needs to happen on my side :)

### DIFF
--- a/integration/rabbitmq/e2e/configuration.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/configuration.e2e-spec.ts
@@ -1,4 +1,8 @@
-import { RabbitMQConfig, RabbitMQModule } from '@golevelup/nestjs-rabbitmq';
+import {
+  RabbitMQConfig,
+  AmqpConnection,
+  RabbitMQModule,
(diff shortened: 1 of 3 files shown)


underfisk (changes requested):
LGTM, i would ask for some test coverage 🙏

frankmangone:
> LGTM, i would ask for some test coverage 🙏

Will see what I can do about that!

frankmangone:
@underfisk Added the requested tests, mind checking if that works for you? Looking good on my end :)

underfisk (approved):

underfisk:
@frankmangone Thanks for this addition, LGTM!

PR #512, opened 2022-10-10
Fix/nest js freeze on rabbit rpc annotation

vijay94:


vijay94:
@WonderPanda Can you verify this PR?

WonderPanda:
@vijay94 Is there a way for us to get some e2e test coverage to verify that this is working as expected? I'm wary of untested changes especially since it seems like the last attempt at fixing this problem didn't actually work

vijay94:
@WonderPanda I can explore on e2e test. But this change is not a tested one. We had two annotations namely RabbitSubscribe and RabbitRPC. The last PR only fixed the RabbitSubscribe annotation. We missed out the RabbitRPC annotation.

vijay94:
> @vijay94 Is there a way for us to get some e2e test coverage to verify that this is working as expected? I'm wary of untested changes especially since it seems like the last attempt at fixing this problem didn't actually work

Hi @WonderPanda Added e2e test to test endpoints when rabbitmq connection is not available.

vijay94:
@underfisk Can we publish a new version with this change?

underfisk:
@vijay94 I don't have permissions to publish but i believe that @WonderPanda will publish a new version soon

WonderPanda:
@vijay94 I published new package versions this morning

PR #521, opened 2022-10-26
fix(ts-jest): unknown property in deep typings issues

regevbr:
The `DeepPartial` type doesn't handle unknown types properly.

Given
```ts
      class Base {
        field?: unknown;
      }

      class Test {
        get base(): Base {
          return undefined as any;
        }
      }

      const base = createMock<Base>();
      const test = createMock<Test>({
        base,
      });
```
You receive the following error:
```
error TS2322: Type 'DeepMocked<Base>' is not assignable to type 'DeepPartial<Base> | undefined'.
      Type 'DeepMocked<Base>' is not assignable to type 'DeepPartial<Base>'.
        Types of property 'field' are incompatible.
          Type 'unknown' is not assignable to type 'DeepPartial<unknown> | undefined'.
```

The fix is easy - stop the recursion when `unknown` type is encountered.

This PR adds the reproducible test and the simple fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,12 @@ name: Continuous Integration
 
 on: pull_request
 
-concurrency: 
+concurrency:
   group: ${{ github.head_ref }}
   cancel-in-progress: true
 
 env:
-  LATEST_NODE_VERSION: 17
+  LATEST_NODE_VERSION: 18
 
 jobs:
   ci:
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [14, 16, 17]
+        node: [14, 16, 18, 19]
       fail-fast: false
 
     services:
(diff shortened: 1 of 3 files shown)


underfisk:
This looks right to me, in case there is `unknown` we shouldn't try to guess to prevent weird outcomes

regevbr:
@underfisk I update the ci to run on node 18 and 19 (instead of EOL 17)

underfisk:
> @underfisk I update the ci to run on node 18 and 19 (instead of EOL 17)

I'm okay with supporting more newer versions

regevbr:
> > @underfisk I update the ci to run on node 18 and 19 (instead of EOL 17)
> 
> I'm okay with supporting more newer versions

Cool! all CIs seems to pass :-)

underfisk (approved):
LGTM

PR #522, opened 2022-10-27
feat(module): allow to configure two secrets for different stripe webhooks types(account & connect)

### LINKED ISSUE #355: Multiple stripe webhooks
Is it possible to set this up to work with two different webhook endpoints? (I need one for connect accounts & one for direct accounts.)

korifey91:
BREAKING CHANGE: Stripe module `webhookConfig`

closes #355


underfisk:
@korifey91 Looks like we have tests failing, we should also reflect the changes on those

underfisk (approved):
LGTM

PR #524, opened 2022-10-29
fix: issue with scheduled events without a comment

foxted:
I recently noticed that Scheduled Events were not being processed anymore, we encountered errors like this: 

```bash
Handler not found for 94605691-b97a-47f2-b03e-3368d4de16bc - {"stack":["HasuraModule"]}
```

Turns out, 3 months ago, this line was introduced, which will try to use the `id` of the event instead of the name, unless a comment has been specified on the event definition: 

https://github.com/golevelup/nestjs/blob/master/packages/hasura/src/hasura.module.ts#L175

I suggest we remove this condition to avoid having to set a comment on scheduled events.


underfisk:
@WonderPanda This aligns with what we have been seeing

WonderPanda:
@foxted Good find, thanks for the contribution

PR #528, opened 2022-11-11
docs(testing): updated ts-jest install docs

adamgronberg:
Updated ts-jest install docs so that is added as a devDependency,

Testing libraries do not need to be added in a production environment. Good to, by default, recommend splitting dev/production dependencies.

### DIFF
--- a/packages/testing/README.md
+++ b/packages/testing/README.md
@@ -27,13 +27,13 @@ For examples we'll show how well it works with NestJS' [ExecutionContext](https:
 Pretty standard installation, nothing too crazy
 
 ```sh
-npm i @golevelup/ts-jest
+npm i @golevelup/ts-jest --save-dev
 ```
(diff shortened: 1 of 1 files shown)


underfisk (approved):
LGTM! Thank you for this change
A minor thing i would probably add is the alias `-D` instead of having to type `--save-dev` or `--dev`

adamgronberg:
I was considering that. But in my head, most people will copy the command.
And if they don't copy it:
* People with experience will know what --save-dev/--dev means and will type the alias themself.
* People who don't know what --save-dev/--dev means will get a bit more explanation of what the command does.
jest has the same way of doing it, for example: https://jestjs.io/docs/getting-started

Will change if you still feel like -D is the way forward!

Love the package btw! Been looking for this for a while :tada:

PR #530, opened 2022-11-15
feat(rabbitmq): return promise from publish to be able to await delivery

### LINKED ISSUE #519: Publish does not send message if applications shuts down after directly
In a test script of us I realised that calling ampqConnection.publish will not deliver messages if there is not at least one await after it.

```ts
public doesNotDeliverMessage() {
  this.amqpConnection.publish(...)
}

public async doesDeliverMessage() {
  this.amqpConnection.publish(...)
  await new Promise(resolve => setTimeout(resolve, 1000))
}
```

is there some way to check if there are not yet delivered messages?

wodka:
fix https://github.com/golevelup/nestjs/issues/519

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -324,12 +324,17 @@ export class AmqpConnection {
       first()
     );
 
-    this.publish(requestOptions.exchange, requestOptions.routingKey, payload, {
-      replyTo: DIRECT_REPLY_QUEUE,
-      correlationId,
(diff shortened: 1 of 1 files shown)


underfisk:
@wodka Could you add some test coverage?
Also it looks like the current tests are broken

wodka:
will later today to improve test coverage and check why the tests are failing

wodka:
this should fix the unit test - the returned promise is beeing tested when ever there is ie a call to request / publish, or how would you like an additional test case?

PR #535, opened 2022-12-01
feat(rabbitmq): add consumeOption to createSubscriber()

m1n-q:


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -351,7 +352,8 @@ export class AmqpConnection {
   public async createSubscriber<T>(
     handler: SubscriberHandler<T>,
     msgOptions: MessageHandlerOptions,
-    originalHandlerName: string
+    originalHandlerName: string,
+    consumeOption?: ConsumeOptions
```
underfisk: Rename to plural `consumeOptions`

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -14,6 +14,15 @@ export interface RabbitMQExchangeConfig {
   options?: Options.AssertExchange;
 }
 
+export interface ConsumeOptions {
```
underfisk: Should this extend or pick from the official library? I wonder if they do have an internal/public interface if we shouldn't use it directly. Could you investigate that?

underfisk (changes requested):
LGTM, i have some minor requests

underfisk (approved):

packages/rabbitmq/src/rabbitmq.interfaces.ts
m1n-q: Oh, sorry. I didn't notice the request because the project using the library was finished before your comment.

PR #561, opened 2023-02-28
ci: drop support for node 14

underfisk:
Drops support for node 14 as it will be out of support on 30 April 2023
We should advise our users to use at least the `LTS` version

## Resources
https://azure.microsoft.com/en-us/updates/community-support-for-node-14-lts-is-ending-on-30-april-2023/
https://nodejs.dev/en/about/releases/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [14, 16, 18, 19]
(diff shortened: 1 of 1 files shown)


underfisk:
@WonderPanda This would require a manual merge as we're removing Node 14 version but it expects to run

underfisk:
@WonderPanda I'm bumping to see if we can drop CI 14, possibly CI 16 as it LTS is dropping out in September

PR #562, opened 2023-02-28
chore: update npm packages

underfisk:
update npm packages to the latest version or next relevant major

Highlights:
- Nestjs goes to the latest
- RxJS goes to the latest
- amqlib gets to `0.10.3` which should give the possibility for new features if devs require something more sophisticated
- dev dependencies get also to the latest such as ts-jest, commit linter, prettier, etc


underfisk:
I'll be splitting into multiple PR's, easier to scope thing out

PR #563, opened 2023-02-28
feat(rabbitmq): upgrades all amqlib to `0.10.3` and `amqp-connection-manager` to `3.9.0`

### LINKED ISSUE #542: Bump amqplib dependency to avoid bluebird
The rabbitmq package adds amqplib^0.8.0 dependency, which is dependent on bluebird. Either use amqplib as peerDependency or update the version please

underfisk:
Safely upgrades both libraries `amqlib` to `0.10.3` and `amqp-connection-manager` to `3.9.0` to keep it up to date even though the latest of `amqp-connection-manager` would be v4 where there are breaking changes

Closes #542


jblyberg:
Is the only thing hold back the merge of this pull request the broken json in package.json? Looks like it's missing a comma. It would be great to upgrade these packages.

PR #587, opened 2023-04-29
fix(testing): fix breaking change from 0.3.6

### LINKED ISSUE #586: testing 0.3.6 breaking change when providing `constructor`
issue #583 that was fixed in #584 is a breaking change - it causes a mock not to equal itself, as Jest checks for constructor equality.
This can be solved by only setting the constructor if it is not already set by the partial object. Currently, there is no way around it.

I will add a PR with some tests

The below passes before 0.3.6

```typescript
export const createComparableMock: typeof createMock = <T extends object>(
  partial?: PartialFuncReturn<T>,
  options?: MockOptions,
): DeepMocked<T> => {
  const _partial: PartialFuncReturn<T> = partial || {};
  const __partial: PartialFuncReturn<T> & { asymmetricMatch: null; __mockName?: string } = {
    asymmetricMatch: null,
    ..._partial,
  };
  if (options?.name) {
    _partial.mockName = __partial._mockName || options.name;
  }
  return createMock<T>(__partial, options);
};

const comparable = createComparableMock<{ foo: number }>({}, { name: 'f1' });
expect([comparable]).toEqual([comparable]);
```

regevbr:
fix constructor not the same for same mock

fix #586

### DIFF
--- a/packages/testing/src/mocks.spec.ts
+++ b/packages/testing/src/mocks.spec.ts
@@ -140,7 +140,7 @@ describe('Mocks', () => {
         base,
       });
 
-      expect(test.base).toBe(base);
+      expect(test.base).toEqual(base);
     });
   });
 
--- a/packages/testing/src/mocks.spec.ts
+++ b/packages/testing/src/mocks.spec.ts
@@ -224,12 +224,30 @@ describe('Mocks', () => {
       expect(result).toBe(42);
     });
 
-    it('should have constructor defined', () => {
-      class Service {}
+    describe(`constructor`, () => {
+      it('should have constructor defined', () => {
+        class Service {}
 
-      const mock = createMock<Service>();
+        const mock = createMock<Service>();
 
-      expect(mock.constructor).toBeDefined();
+        expect(mock.constructor).toBeDefined();
+      });
+
+      it('should have the same constructor defined', () => {
+        class Service {}
+
+        const mock = createMock<Service>();
+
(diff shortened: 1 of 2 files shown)


regevbr:
@WonderPanda @mentos1386 can you please have a look? You introduced a breaking change in 0.3.6

mentos1386 (approved):
LGTM, thanks for catching this!

You can also revert the [L143](https://github.com/golevelup/nestjs/commit/e31d2b39c4647fc9e96afa9e68ad41f244e6fc20#diff-0e5e195e64de75fc7fd3b8171c1109cd90cfbb99786ace3ec28c730fbe1c7b99R143) in `mocks.specs.ts` that i made. I didn't realize it's a bigger issue at the time :(

regevbr:
> LGTM, thanks for catching this!
> 
> You can also revert the [L143](https://github.com/golevelup/nestjs/commit/e31d2b39c4647fc9e96afa9e68ad41f244e6fc20#diff-0e5e195e64de75fc7fd3b8171c1109cd90cfbb99786ace3ec28c730fbe1c7b99R143) in `mocks.specs.ts` that i made. I didn't realize it's a bigger issue at the time :(

Changed - didn't even notice it was changed by you. This line is exactly the test that should have raised a red flag :-)

WonderPanda (approved):

WonderPanda:
Thanks! This is available in @golevelup/ts-jest@0.3.7

PR #601, opened 2023-06-03
chore(rabbitmq): amqp libs update

### LINKED ISSUE #542: Bump amqplib dependency to avoid bluebird

witem:
Update `amqplib` and `amqp-connection-manager` to the latest version.
There are many fixes and improvements.

Similar to https://github.com/golevelup/nestjs/pull/459 and https://github.com/golevelup/nestjs/pull/563
Will close #542 

BREAKING CHANGES
We will no longer emit a disconnect event on an 
initial connection failure - instead we now emit connectFailed on each
connection failure, and only emit disconnect when we transition from
connected to disconnected.

### DIFF
--- a/packages/rabbitmq/package.json
+++ b/packages/rabbitmq/package.json
@@ -37,12 +37,11 @@
     "@golevelup/nestjs-common": "^1.4.4",
     "@golevelup/nestjs-discovery": "^3.0.0",
     "@golevelup/nestjs-modules": "^0.6.1",
-    "amqp-connection-manager": "^3.0.0",
-    "amqplib": "^0.8.0"
+    "amqp-connection-manager": "^4.1.13",
+    "amqplib": "^0.10.3"
   },
   "devDependencies": {
-    "@types/amqp-connection-manager": "^2.0.4",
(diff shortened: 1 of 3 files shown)


underfisk:
I'm down for pushing this but it would have to be a major release as we're adding breaking changes. @WonderPanda

jarpoole:
Documented another key bug that would be fixed by upgrading `amqp-connection-manager` to v4 in https://github.com/golevelup/nestjs/issues/647. If there is still work to be done besides the major version change I'm happy to help out!

underfisk:
> Documented another key bug that would be fixed by upgrading `amqp-connection-manager` to v4 in #647. If there is still work to be done besides the major version change I'm happy to help out!

Feel free to help this branch get across the finish line, if the merge conflicts get resolved and the test updated (if needed) I'll gladly push this and release it as part of the next major version

ttshivers:
Why was this closed?

PR #608, opened 2023-06-21
fixed the order of generated metadata

ali-shabani:
### Description:
This package creates the metadata for events like this:
```yaml
num_retries: 3
interval_sec: 15
```

But when you export the metadata from hasura, the order is different:
```yaml
interval_sec: 15
num_retries: 3
```
This means you have to commit those changes every time if the order of exporting and generating metadata is not consistent. For example, if I change something in hasura and export the metadata and commit those changes, the next time I change my code, the order of the generated metadata will change and I have to commit those files again.

The manual solution is to make sure the order of generating/exporting metadata is the same, but that's annoying when you can fix this by changing one line of code.


underfisk (approved):
lgtm

Issue #567, opened 2023-03-06
Routing key is not considered in handlers


bintoll:
I am still new in rabbitmq, so please correct me if I am wrong. I feel quite odd behavior having a multiple rpc handlers over the same queue and exchange. I have read about the possible combinations of queues, exchanges, and routing keys and looks like this is a correct setup. 
I configured rabbitmq module with exchange with name "amq.topic" and type "topic". I also setup "default" handler, where I set the queue name and default exchange name. (I think it could be potentially implemented a little bit easier for new comers as it is in the nestjs/microserves package, all defaults are set by the library itself. But that's another topic.) 
The connection works, messages are delivered. Then I added some more rpc handlers in controller. After this I see odd behavior that the message that were send travels from one rpc handler to another one by one until one of them succeeds. That is really strange behavior. After digging in source code I figured out that in setupQueue method all possible routing keys are bind to the same queue which is correct. After I checked how these messages are handled: in amqp/connection file in setupRpcChannel method in channel.consume callback function I get all messages incoming for this queue with all possible routingKeys. In this callback after await this.handleMessage throws an error it is catched and then the message is REQUEUE (behavior by default). And in this case it goes to next rpc handler until success. As I see the routingKey parameter is not considered at all and is meaningless. After implementing a quick and dirty patch it works as expected, the correct rpc handler with same routingKey handles the message.

<img width="968" alt="Снимок экрана 2023-03-06 в 11 39 39" src="https://user-images.githubusercontent.com/20185572/223014388-42fe199a-e972-4955-b997-ff724b2ef2ce.png">

Could someone please comment this. Is it a bug or there is a misunderstanding from my side? Why routingKey is not considered at all in current implementation?

youngcholo:
I encountered the same behaviour mentioned above in my current RabbitMQ setup.

```
  @RabbitRPC({
    exchange: 'rpcExchange',
    routingKey: 'rpc.test.item.create',
    queue: 'test-rpcExchange',
  })
  async handleCreate(data: JSON): Promise<string> {
    this.items.push(data);
    console.log(`Created item: ${JSON.stringify(data)}`);
    return 'Item created successfully';
  }

  @RabbitRPC({
    exchange: 'rpcExchange',
    routingKey: 'rpc.test.item.findAll',
    queue: 'test-rpcExchange',
  })
  async handleFindAll(): Promise<string> {
    console.log('FindAll items', this.items);
    return JSON.stringify(this.items);
  }
```
[...]

PR #615, opened 2023-07-04
Update connection.ts

eabucejo:
Nack it if the msg.fields.rountingKey is not equal to the rpcOptions.routingKey.

Fix to : https://github.com/golevelup/nestjs/issues/567

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -484,6 +484,11 @@ export class AmqpConnection {
             throw new Error('Received null message');
           }
 
+          if (rpcOptions.routingKey !== msg.fields.routingKey) {
+                    channel.nack(msg, false, true);
+                    return;
+          }
+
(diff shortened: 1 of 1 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -484,6 +484,11 @@ export class AmqpConnection {
             throw new Error('Received null message');
           }
 
+          if (rpcOptions.routingKey !== msg.fields.routingKey) {
```
underfisk, line 487: Can you just run prettier? It has alot of extra tab spacing

underfisk (changes requested):
Lgtm just a formatting issue

packages/rabbitmq/src/amqp/connection.ts
eabucejo: Fixed formatting issue. Cheers!

underfisk (approved):

PR #630, opened 2023-08-04
docs: use rawBody instead of webhooks lib as the default

unix:
Now we don't need to disable `json` parsing with an additional module, we can do it by accessing the `rawBody` that comes with nestjs.

This PR just changes the documentation, nothing changes.


underfisk:
@unix Thanks for the contribution!

PR #636, opened 2023-08-20
fix(rabbitmq): reading logger config as undefined property in AmqpCon…

VarTony:
Hi.
When accessing the config's property logger, the following error occurred:
`TypeError: Cannot read properties of undefined (reading 'logger')
    at new AmqpConnection (/Users/.../auth-service/node_modules/@golevelup/nestjs-rabbitmq/src/amqp/connection.ts:132:22)`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,22920 @@
+{
+  "name": "root",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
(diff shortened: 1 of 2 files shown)


AckerApple:
What did you do to get past this issue? Did you find some docs or something? Couple of us trying to figure it out in #871

PR #642, opened 2023-09-01
fix: apply the rountingKey fix for subscriptions as well

### LINKED ISSUE #567: Routing key is not considered in handlers

asanovr:
Similar https://github.com/golevelup/nestjs/pull/615 but for subscriptions
fix #567


asanovr:
Looks like I have misspublished, Should I revert the last commit?

underfisk:
> Looks like I have misspublished, Should I revert the last commit?

Yes you should revert that

asanovr:
@underfisk Done

underfisk:
@asanovr Can you make sure the test coverage is updated or passing locally? It seems to be failling

underfisk:
@asanovr Are you able to fix the test coverage? If so I'll get this merged asap

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -393,6 +393,11 @@ export class AmqpConnection {
             throw new Error('Received null message');
           }
 
+          if (msgOptions.routingKey !== msg.fields.routingKey) {
+            channel.nack(msg, false, true);
```
Dryymoon, line 397: This will cause silent infinity loop without any notifications in case any configuration differences beetwen rabbit and routingKey definitions, thant not good. I recommened here - throw some message for handle las in custom errorHandler by this lib or for handle by default error behaviour.

PR #648, opened 2023-09-22
fix(rabbitmq): using `__routeArguments__` key to allow pipe injection

### LINKED ISSUE #645: Pipe not working with @RabbitPayload decorator
Hi,

The pipe I'm adding to my payload is not being executed:
```ts
protected async onBridgeRequest(
  @RabbitPayload(ParsePayloadPipe) payload: Payload) {
  console.log({ payload }) // payload is just the raw, un-transformed data
}
```
Pipes only execute if added with `@UsePipes()` at the handler, but then it applies the pipe to all deocrated arguments. I also use `@RabbitHeader()` to get a header value, and the pipe attempts to run on it as well (which I guess is expected):
```ts
protected async onBridgeRequest(
  @RabbitPayload(ParsePayloadPipe) payload: Payload, @RabbitHeader('foo') foo: string) {
  console.log({ payload })
}
```

So now I just add the pipe to the handler, and check in `transform` if the `argument.metatype` matches `Payload` then I execute the pipe logic, otherwise, I just send the data through as-is. It's very hacky, but the only way I can get this working as desired.

Not sure why the pipe(s) passed through that decorator don't work, as it's part of the NestJS contract with parameter decorators (first data, then pipe(s)).

sezanzeb:
Fixes https://github.com/golevelup/nestjs/issues/645

nestjs regular `@Param` decorator and such adds the pipes to the metadata (via the reflection methods) using the `'__routeArguments__'` key (https://github.com/nestjs/nest/blob/93b99d09ed84640815fc3aa99196cf94e692bc4d/packages/common/decorators/http/route-params.decorator.ts#L80)

The `DependenciesScanner` of nestjs then loads the `'__routeArguments__'` and injects the pipes in `reflectParamInjectables` (https://github.com/nestjs/nest/blob/master/packages/core/scanner.ts)

golevelup/nestjs however uses the `'RABBIT_ARGS_METADATA'` key for it (https://github.com/golevelup/nestjs/blob/c927cb12f5203fe4739027fc54d78c6ae2629cfb/packages/rabbitmq/src/rabbitmq.decorators.ts), so the scanner will never load and inject them.

I just changed it to `'__routeArguments__'`, and then pipes were correctly created, and their dependencies injected. I don't know enough about the internals to judge if this causes other side-effects. Hopefully tests will catch it all. I wasn't able to run tests locally, so I'm waiting for the ci pipeline to run.

### DIFF
--- a/packages/rabbitmq/src/rabbitmq.constants.ts
+++ b/packages/rabbitmq/src/rabbitmq.constants.ts
@@ -1,6 +1,5 @@
 export const RABBIT_HANDLER = Symbol('RABBIT_HANDLER');
 export const RABBIT_CONFIG_TOKEN = Symbol('RABBIT_CONFIG');
-export const RABBIT_ARGS_METADATA = 'RABBIT_ARGS_METADATA';
 export const RABBIT_PARAM_TYPE = 3;
 export const RABBIT_HEADER_TYPE = 4;
 export const RABBIT_REQUEST_TYPE = 5;
--- a/packages/rabbitmq/src/rabbitmq.decorators.ts
+++ b/packages/rabbitmq/src/rabbitmq.decorators.ts
@@ -7,9 +7,9 @@ import {
   Type,
   assignMetadata,
 } from '@nestjs/common';
+import { ROUTE_ARGS_METADATA } from '@nestjs/common/constants';
 import { isString } from 'lodash';
 import {
-  RABBIT_ARGS_METADATA,
   RABBIT_CONFIG_TOKEN,
   RABBIT_HANDLER,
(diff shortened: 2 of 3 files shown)


underfisk:
Thank you for pushing this PR. 
I did think about this before and I believe it was intentional not to have the same key so as not to confuse with controllers. Unfortunately I've been wanting to refactor the package as RabbitMQ providers should have been controllers in the first place but if you're change doesn't introduce any breaking change, I'll be happy to merge

sezanzeb:
At least I haven't observed any problems with this until now. I'll try working with this a bit more over the next one or two weeks, and report back if there have been any unexpected bugs.

Issue #659, opened 2023-11-04
[RabbitMQ] create queues without consumer


fermentfan:
Hi,
I am currently trying to send out messages with a delay by leveraging the TTL + DLQ pattern as described here:

https://github.com/golevelup/nestjs/issues/138#issuecomment-853936061

At the moment I solely rely on queue creation by this dependency, which I really love! But to follow the pattern one must create a queue without any handler added to it where the messages just wait until their TTL hit. Sadly I think this is not possible? I'd have expected that one could define queues with options like DLQ next to the exchanges in the module definition. It would be so cool if that was possible :)

WonderPanda:
It would honestly be pretty easy to get this functionality implemented but I'll be honest I haven't used RabbitMQ in so many years now that I have no motivation. I bet you it could be a really easy contribution for you though and you'll get your face here: https://github.com/golevelup/nestjs#contributors-

papalardo:
My config:

```js

export const dlq = (queue: string) => queue + '.dlq';
export const dlx = (exchange: string) => exchange + '.dlx';

@Global()
@Module({
  providers: [],
  imports: [
    RabbitMQModule.forRootAsync(RabbitMQModule, {
      useFactory: () => ({
        uri,
        exchanges: [
          {
            name: RMQ_EXCHANGES.WORKERS,
            type: 'direct',
          },
          {
            name: dlx(RMQ_EXCHANGES.WORKERS),
            type: 'direct',
          },
        ],
        channels: {
          [RMQ_QUEUES.CAMPAIGN_DISPATCHED]: {
            prefetchCount: 1,
          },
        },
        defaultSubscribeErrorBehavior: MessageHandlerErrorBehavior.NACK,
      }),
    }),
  ],
  exports: [RabbitMQModule],
})
export class RabbitMqModule implements OnModuleInit {
  constructor(private readonly amqpConnection: AmqpConnection) {}

  async onModuleInit() {
    await this.amqpConnection.channel.assertQueue(dlq(RMQ_QUEUES.CAMPAIGN_DISPATCHED), {
      durable: true,
    });

    await this.amqpConnection.channel.bindQueue(
      dlq(RMQ_QUEUES.CAMPAIGN_DISPATCHED),
      dlx(RMQ_EXCHANGES.WORKERS),
      dlq(RMQ_QUEUES.CAMPAIGN_DISPATCHED),
    );
  }
}
```

PR #662, opened 2023-11-06
Feat/659 unbound queues

fermentfan:
This introduces unbound queues to be added to the module definition like described in #659. I hope I found everything important :)

### DIFF
--- a/integration/rabbitmq/e2e/configuration.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/configuration.e2e-spec.ts
@@ -16,6 +16,7 @@ const amqplibUri = `${uri}?heartbeat=5`;
 const logger = new ConsoleLogger('Custom logger');
 
 const nonExistingExchange = 'non-existing-exchange';
+const nonExistingQueue = 'non-existing-queue';
 
 class RabbitConfig {
   createModuleConfig(): RabbitMQConfig {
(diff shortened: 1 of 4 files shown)


integration/rabbitmq/e2e/configuration.e2e-spec.ts
```diff
@@ -160,6 +161,72 @@ describe('Module Configuration', () => {
...
+                  wait: true,
+                  reject: true,
+                  timeout: 3000,
+                },
+                logger,
+              }),
+            ],
+          }).compile();
+
+          fail(
```
WonderPanda, line 187: It looks like you're forcing the test to throw an error here. Should this be based on the internals and having checkQueue ultimately fail?

WonderPanda (changes requested):

integration/rabbitmq/e2e/configuration.e2e-spec.ts
fermentfan: tbh I just copied the exchange [test](https://github.com/golevelup/nestjs/blob/da3545241e987e8f5bcd8d275aa94d049456833b/integration/rabbitmq/e2e/configuration.e2e-spec.ts#L86) :D

fermentfan: Looks like this is a workaround for asserting that the line was never hit, because it should always fail in beforehand.

fermentfan:
Should run green now!

integration/rabbitmq/e2e/configuration.e2e-spec.ts
WonderPanda: Hehe thanks for the reminder. Its been a long time since I looked at the exchange code and forgot the approach that was taken there

fermentfan:
Just tried it out and spotted that I am missing the ability to bind it to an exchange like this. Gonna take another look on how to do that.

fermentfan:
@WonderPanda the queues are also bindable to an exchange with routing keys. I also already tested it locally! Would be glad about a review.

fermentfan:
@underfisk do you also have review & merge rights?

underfisk:
> @underfisk do you also have review & merge rights?

Hi, i do have the review & merge rights. As soon as i have some free time i'll try to switch context and review the PR, meawhile i'll approve for CI to re-run

underfisk (approved):
In general LGTM, it looks like a nice feature to have. 
The types could be improved tho, for example you could have the `RabitMQQueConfig` being built on top of the underlying types instead of using any types but it could be a follow up improvement

WonderPanda:
@fermentfan This should be available in `@golevelup/nestjs-rabbitmq@4.1.0`

Please try it out and let us know if its working for you

PR #663, opened 2023-11-17
fix: use the Logger passed in RabbitMQModule config instead of default

lucas-gregoire:
Hello, I discovered an issue when passing a custom Logger to the `RabbitMQModule`:

image

Some logs are made using the logger passed in the config:

image
image

But some other logs, like ones written in `onApplicationBootstrap`, don't use the Logger provided in the config but the `logger` property of the `RabbitMQModule` which is not configurable.

image

----

So my fix consists of using the logger from the provided config everywhere in the `RabbitMQModule`.
And now all logs are correct:

image


underfisk:
@lucas-gregoire Thank you for the fix, LGTM i'll see if test passes and then have it merged

PR #668, opened 2023-12-10
feat: add wildcard topic matcher, support * and # in routingKeys

Dryymoon:
Thanks for a good library!

This fix add:
* Support for amq.topic willcard routing key matching using '@artie-owlet/amqp-routing-match' lib

For example: message with routing key `service.v1.log` should match routing key `service.v1.*` and `service.#`

* Fixed silently message looping when getting message with mismatching with defined routing key.

* Fixed package.json mismatch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage/
 tsconfig.build.tsbuildinfo
 yarn-error.log
 integration/rabbitmq/dist
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ coverage/
 tsconfig.build.tsbuildinfo
 yarn-error.log
 integration/rabbitmq/dist
-package-lock.json
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -548,8 +548,9 @@ export class AmqpConnection {
               topicPattern.match(msg.fields.routingKey || '')
             )
           ) {
-            channel.nack(msg, false, true);
-            return;
+            throw new Error(
+              'Received message with mismatch defined routing key or keys'
+            );
           }
 
           const response = await this.handleMessage(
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -7,4 +7,5 @@ packages/testing/*/lib
 coverage/
 tsconfig.build.tsbuildinfo
 yarn-error.log
-integration/rabbitmq/dist
\ No newline at end of file
+integration/rabbitmq/dist
+package-lock.json
```
underfisk: `package-lock.json` must not be ignored, if you remove this change i'll be able to change

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -524,6 +525,15 @@ export class AmqpConnection {
   ): Promise<ConsumerTag> {
     const queue = await this.setupQueue(rpcOptions, channel);
 
+    let topicPatterns: TopicPattern[];
```
underfisk, line 528: This looks like a new pattern to use using this library, would you be able to develop some unit tests around these changes? This will definitely change quite a lot

.gitignore
Dryymoon: Resolved

underfisk (approved):

underfisk:
@Dryymoon Could you rebase?

underfisk:
@ttshivers If you have some free time and could help the buddy out, i can probably get this merged soon. I wish there was some test coverage as I've never used this library before, have you ever used it? Are there any cons of getting it merged using this library?

ttshivers:
> @ttshivers If you have some free time and could help the buddy out, i can probably get this merged soon. I wish there was some test coverage as I've never used this library before, have you ever used it? Are there any cons of getting it merged using this library?

I'm trying to understand the goal of this PR. Is this to double check that the message received on a queue matches the routing keys defined in code? Is the thought that there might be extra queue bindings defined on rabbitmq and we should do consumer verification and throw errors when we get unexpected messages?

underfisk:
I'm closing this PR, we'll re-open if we figure out that it is actually valuable. @ttshivers has left some interesting thoughts, i'd love to also understand a bit better this change

Issue #667, opened 2023-12-08
Initializing Rabbit Module creates random queues


shreyassood:
On initializing a module with a queue specified, the library creates random queues if `createQueueIfNotExists` is on.

This happens at https://github.com/golevelup/nestjs/blob/be88a40d5b4c988d2b5e855b446f7a8e3b43b9bf/packages/rabbitmq/src/amqp/connection.ts#L304-L305 within the setupQueuesWithBindings method. 

The setupQueue method expects a subscriptionsOptions object of type `MessageHandlerOptions` but is passed in an object of type `RabbitMQQueueConfig`. This object does not have the queue property https://github.com/golevelup/nestjs/blob/be88a40d5b4c988d2b5e855b446f7a8e3b43b9bf/packages/rabbitmq/src/amqp/connection.ts#L671-L672 and causes Rabbit to initialize a queue with a [random name] (https://www.rabbitmq.com/queues.html#server-named-queues).

<img width="964" alt="Screenshot 2023-12-08 at 5 45 28 PM" src="https://github.com/golevelup/nestjs/assets/2962613/9ca0b86c-a724-4dc3-ab07-bf606e1a046f">

Also on a side note: should assertQueue not create a queue at https://github.com/golevelup/nestjs/blob/be88a40d5b4c988d2b5e855b446f7a8e3b43b9bf/packages/rabbitmq/src/amqp/connection.ts#L304-L305 regardless anyway (`createQueueIfNotExists`) in this case shouldn't matter?

liudichen:
I also encountered this issue, and it's not just the queue attribute, but also the queueOptions attribute with different names in the two configurations

maybe should modify this line:
https://github.com/golevelup/nestjs/blob/be88a40d5b4c988d2b5e855b446f7a8e3b43b9bf/packages/rabbitmq/src/amqp/connection.ts#L302

to
```
this.setupQueue({...configuredQueue,queue: configuredQueue.name,queueOptions:configuredQueue.options}, channel)
```

underfisk:
The next version shall include the fix for this, bumping here for @WonderPanda

huantaoliu:
nice

shreyassood:
Awesome, thanks for working on this!

underfisk:
> Awesome, thanks for working on this!

All the merit goes to @ttshivers ! He's doing an amazing job helping out

trickaugusto:
Hello everyone!

[...]

PR #671, opened 2024-01-08
fix issues#667

liudichen:
Attempt to fix[ issue # 667:](https://github.com/golevelup/nestjs/issues/667)

At present, the queue name attribute corresponds to the queue attribute in Message HandlerOptions, while it corresponds to the name in RabbitMQQueueConfig. There is also an issue of inconsistent queueOptions and options attribute names.

This can cause RabbitMQ to create many anonymous queues (unable to consume and automatically delete) due to no queue in RabbitMQQueueConfig, or channel initialization errors due to different configurations of the same queue at different locations due to the lack of queueOptions.

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -299,7 +299,7 @@ export class AmqpConnection {
       const { createQueueIfNotExists = true } = configuredQueue;
 
       if (createQueueIfNotExists) {
-        this.setupQueue(configuredQueue, channel);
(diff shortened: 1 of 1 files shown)


underfisk:
@liudichen Thank you for the contribution! Would you be able to develop some test coverage to assert that this functionality works as intended?

underfisk:
@ttshivers Is this still applicable or shall i close it?

ttshivers:
This shouldn't be needed anymore since my fix merged at https://github.com/golevelup/nestjs/pull/676

PR #674, opened 2024-01-15
fix(rabbitmq): fix unawaited promises

ttshivers:
I noticed some floating promises in the code when trying to debug an issue due to ordering. Ensuring all the promises are awaited where expected should help prevent possible issues like ensureing the assertExchange is done before trying to bind a queue to it.

### DIFF
--- a/integration/rabbitmq/e2e/configuration.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/configuration.e2e-spec.ts
@@ -18,6 +18,8 @@ const logger = new ConsoleLogger('Custom logger');
 const nonExistingExchange = 'non-existing-exchange';
 const nonExistingQueue = 'non-existing-queue';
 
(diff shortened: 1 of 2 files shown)


underfisk:
@ttshivers Thank you for this change! Would you be down to write some test coverage to assert what you're describing/what we're trying to accomplish here?

ttshivers:
> @ttshivers Thank you for this change! Would you be down to write some test coverage to assert what you're describing/what we're trying to accomplish here?

I have written a test that fails before my fix and passes after my fix which exposes one of the race conditions that is possible by not awaiting the promises.

underfisk (approved):

underfisk:
@ttshivers could you update the branch and fix conflicts?

ttshivers:
> @ttshivers could you update the branch and fix conflicts?

I have resolved the conflicts

PR #676, opened 2024-01-16
fix(rabbitmq): fix library asserting queues with empty names

ttshivers:
This is an attempt to fix the issue where this library doesn't pass the right arguments to setupQueue in setupQueuesWithBindings. In addition, this removes the duplicate setup that setupQueuesWithBindings was doing that setupQueue alredy does.

re #667


underfisk (approved):
Lgtm, ty

PR #677, opened 2024-01-16
build(rabbitmq): update amqp libs

ttshivers:
Update amqplib and amqp-connection-manager to the latest version. There are many fixes and improvements.

BREAKING CHANGE: We will no longer emit a disconnect event on an initial connection failure - instead we now emit connectFailed on each connection failure, and only emit disconnect when we transition from connected to disconnected.

re #542
re #459


underfisk:
@ttshivers Thank you very much for reviving this

@WonderPanda The next rabbitmq package version should be a major because this implies breaking changes

Issue #673, opened 2024-01-15
RabbitMQ publish doesn't take advantage of amqp-connection-manager's reliable publish


ttshivers:
The rabbitmq package's `publish` implementation calls `publish` on a raw `ConfirmChannel` instead of the `ChannelWrapper` provided by  `amqp-connection-manager`.

Since it's not using the `ChannelWrapper`, it's missing out on all the reliability features that `amqp-connection-manager` provides for publishing messages.

Was this an intentional decision to not use the `ChannelWrapper` ?

https://github.com/golevelup/nestjs/blob/1891b3286cb620d809e9d00f9a56d7805b846607/packages/rabbitmq/src/amqp/connection.ts#L612-L627

Instead of using `this._channel` there, `this._managedChannel` could be used instead which is the `ChannelWrapper`

underfisk:
@ttshivers If you feel confident about this change and can provide test coverage to assert that it works as intended I'll be happy to review and merge. 
I don't know what the reason was behind not to use `ChannelWrapper` but from your explanation it seems logical to use so

ttshivers:
Fixed by https://github.com/golevelup/nestjs/pull/678

PR #678, opened 2024-01-16
feat(rabbitmq): publish using ChannelWrapper

ttshivers:
Publish messages using the ChannelWrapper instead of the raw ConfirmChannel to take advantage of the reliability features of  `amqp-connection-manager`.

BREAKING CHANGE: This changes the behavior of throwing connection related errors

re #673

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -11,7 +11,6 @@ import {
   ConfirmChannel,
   Options,
 } from 'amqplib';
-import { Replies } from 'amqplib/properties';
 import {
   EMPTY,
   interval,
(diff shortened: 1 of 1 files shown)


underfisk:
@ttshivers I love this change, reduces quite some boilerplate. My only concern is if this will work as intended for existing applications. If our test coverage is enough then I'm happy to include this as part of the next major with amq v4

ttshivers:
I looked over the test coverage and it seems sufficient to ensure that messages are still getting published properly now using the channel wrapper instead.

underfisk:
> I looked over the test coverage and it seems sufficient to ensure that messages are still getting published properly now using the channel wrapper instead.

Alright, thanks!

PR #684, opened 2024-01-24
feat(rabbitmq): add persistent reply to

ilyalatt:


.gitignore
```diff
@@ -6,5 +6,6 @@ packages/testing/*/lib
 .idea
 coverage/
 tsconfig.build.tsbuildinfo
+.yarn
```
underfisk: Is there a motive for `.yarn` to be included as part of `.gitignore?`

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -95,6 +95,7 @@ export interface MessageHandlerOptions {
   assertQueueErrorHandler?: AssertQueueErrorHandler;
   allowNonJsonMessages?: boolean;
   createQueueIfNotExists?: boolean;
+  usePersistentReplyTo?: boolean;
```
underfisk, line 103: I would add here some jsdoc with `@default false` because we should be transparent on those
If applicable, have a quick summary of this flag

underfisk (changes requested):
There's just some small things but in general i like this new change.
If possible i would say to create some test coverage as it is useful for us to guarantee that existing consumers of this library do not receive a breaking change (if persistent is false by default they never will but just in case this gives us more confidence)

.gitignore
ilyalatt: Yes, yarn 3 created `.yarn` directory

packages/rabbitmq/src/rabbitmq.interfaces.ts
ilyalatt: I've added the documentation

.gitignore
underfisk: @ilyalatt You should use yarn 2 with this repo :/ we do not intend to use yarn 3, in fact the intention is to switch to PNPM

ilyalatt: Ok, I have reverted this change

ilyalatt:
> There's just some small things but in general i like this new change. If possible i would say to create some test coverage as it is useful for us to guarantee that existing consumers of this library do not receive a breaking change (if persistent is false by default they never will but just in case this gives us more confidence)

I wanted to make a test to be sure. However, now I prefer to avoid it if possible, since running tests is not trivial for me, and I do not want to spend time on such a small change. Can we merge the PR as-is?

Issue #689, opened 2024-01-27
Support Module register with undefined config:  RabbitMQModule.forRoot(RabbitMQModule, undefined)


huantaoliu:
I have a usecase where my rabbitmq config is dynamic,  I want register  Rabbitmq Module with undefined config 
```
RabbitMQModule.forRoot(RabbitMQModule, undefined)
```
Latter in my code, I will use ``connectionManager`` myself to init connection and register handler.

Currently it forces me to use a config when importing Rabbitmq module, and `url` is mandary in ``RabbitMQConfig``, And this ``registerHandlers: false`` still requires us provide rabbitmqconfig at module import time. 

If aggreed, I want to support: ``RabbitMQModule.forRoot(RabbitMQModule, undefined)``

underfisk:
@huantaoliu I'm okay having an optional config as long as it doesn't break any existing feature. If you're willing to push a PR i'll be glad to review.
I don't think we need the explicit undefined but feel free to propose something.

PR #690, opened 2024-01-28
Feature/optional config

huantaoliu:
Support  #689 

```
RabbitMQModule.forRoot(RabbitMQModule)
```

### DIFF
--- a/packages/rabbitmq/src/rabbitmq.module.ts
+++ b/packages/rabbitmq/src/rabbitmq.module.ts
@@ -1,6 +1,5 @@
 import {
   DiscoveredMethod,
-  DiscoveredMethodWithMeta,
   DiscoveryModule,
   DiscoveryService,
 } from '@golevelup/nestjs-discovery';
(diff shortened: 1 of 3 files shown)


packages/rabbitmq/src/rabbitmq.module.ts
```diff
@@ -81,20 +80,28 @@ export class RabbitMQModule
...
-    this.logger = config.logger || new Logger(RabbitMQModule.name);
+    this.logger = config?.logger || new Logger(RabbitMQModule.name);
   }
 
-  static async AmqpConnectionFactory(config: RabbitMQConfig) {
+  static async AmqpConnectionFactory(
+    config: RabbitMQConfig
+  ): Promise<AmqpConnection | undefined> {
+    const logger = config?.logger || new Logger(RabbitMQModule.name);
+    if (config == undefined) {
```
underfisk, line 90: possibly just do `if (!config)`, you don't need to explicitly check for undefined

underfisk (approved):

PR #695, opened 2024-02-11
build(rabbitmq): allow using reflect-metadata 0.2.0

ttshivers:
Allow using reflect-metadata 0.2.0 by updating the allow versions in the peer dependencies

### DIFF
--- a/packages/rabbitmq/package.json
+++ b/packages/rabbitmq/package.json
@@ -47,7 +47,7 @@
   "peerDependencies": {
     "@nestjs/common": "^10.x",
     "@nestjs/core": "^10.x",
-    "reflect-metadata": "^0.1.0",
+    "reflect-metadata": "^0.1.0 || ^0.2.0",
     "rxjs": "^7.x"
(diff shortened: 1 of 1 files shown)


underfisk:
@ttshivers Shall we maybe just bump to 2.0.0?

ttshivers:
> @ttshivers Shall we maybe just bump to 2.0.0?

It doesn't look like there is a version 2.0.0 yet.
https://www.npmjs.com/package/reflect-metadata?activeTab=versions

ttshivers:
> @ttshivers Shall we maybe just bump to 2.0.0?

Or if you meant bumping it to `0.2.0`, I would probably not recommend that yet because not all packages have updated their peer deps or dependencies to 0.2.0 yet and could lead to a situation where no version makes every package happy.

This is also the strategy nestjs uses:
https://github.com/nestjs/nest/blob/d658942e1b8d6d214731297a36d014b330cf0aed/packages/core/package.json#L46

PR #711, opened 2024-03-21
feat(rpc): handle parallel RPCs response

aeza8i:
Hello
There is a race condition between multiple RPCs in concurrency. This is obvious in distributed systems and happens when multiple RPCs use the same correlation id. To avoid this issue, using a request id that is defined in the request header will help.

### DIFF
--- a/integration/rabbitmq/e2e/rpc.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/rpc.e2e-spec.ts
@@ -2,6 +2,7 @@ import { AmqpConnection } from '@golevelup/nestjs-rabbitmq';
 import { INestApplication } from '@nestjs/common';
 import { Test } from '@nestjs/testing';
 import { AppModule } from '../src/app.module';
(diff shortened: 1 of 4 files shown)


underfisk:
@farhadiis Thank you for this change! Would it be valuable to add an example to our documentation on how to leverage `X-Request-Id` header as part of our module? 
I do see value on driving awareness that such feature exist for this particular use case

aeza8i:
@underfisk Thanks for your hint. document updated.

shahrooz1981 (commented):
I've had the opportunity to review the changes made in this PR and I'm pleased to say that everything looks great to me. The code is clean, well-structured, and adheres to the project's standards. 

One aspect that particularly stands out is the consideration for reverse compatibility. This is a crucial factor in maintaining the robustness of the software while ensuring that updates do not disrupt the functionality for existing users. The way this PR handles it is commendable.

I appreciate the effort put into this contribution. It's clear that this will positively impact the project. Keep up the good work!

PR #713, opened 2024-03-24
fix(rabbitmq): a fix for routing key check. The check should respect wildcards

### LINKED ISSUE #712: Wildcard routing stopped working. RabbitMQ
https://github.com/golevelup/nestjs/issues/567
 
Regarding the issue above. It seems that it has broken some existing functionality.

Let me give an example: 
`
@RabbitRPC({
    routingKey: "USER.UPDATED.#",
    queue: "someQueue"
  })
  async handleUserUpdated()
` 
if the message is pushed with routingKey "USER.UPDATED.TEST"  the handler will never be hit. 
The question is, should the wildcards be respected at that point, or is my initial implementation wrong?

PR added
@WonderPanda

Andrius0124:
Added logic to respect wildcards for the routing keys check
fix #712

### DIFF
--- a/packages/rabbitmq/src/amqp/utils.ts
+++ b/packages/rabbitmq/src/amqp/utils.ts
@@ -1,0 +1,20 @@
+export function matchesRoutingKey(
+  routingKey: string,
+  pattern: string[] | string | undefined
+): boolean {
+  if (!pattern) return false;
+  const patterns = Array.isArray(pattern) ? pattern : [pattern];
+  for (const pattern of patterns) {
+    const splitKey = routingKey.split('.');
+    const splitPattern = pattern.split('.');
+
+    for (let i = 0; i < splitPattern.length; i++) {
+      if (splitPattern[i] === '#') return true;
+      if (splitPattern[i] !== '*' && splitPattern[i] !== splitKey[i]) break;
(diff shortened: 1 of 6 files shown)


packages/rabbitmq/src/amqp/utils.ts
```diff
@@ -0,0 +1,20 @@
+export function matchesRoutingKey(
```
underfisk, line 1: i would love to see some test coverage added. 
E2E is also a plus as i'd like to assert that this logic works as expected

underfisk (changes requested):

packages/rabbitmq/src/amqp/utils.ts
Andrius0124: will add some later today

underfisk: @Andrius0124 Alright, just let me know and I'll review/merge if needed

Andrius0124: @underfisk  Updated. Let me know if I missed anything

underfisk (approved):
lgtm, thanks for adding the test coverage

PR #705, opened 2024-03-02
chore: bump stripe version to 14.19.0

fermentfan:
I noticed my local build fails due to the version conflict. I didn't try it out locally yet so I wonder if there is more work involved than just bumping it?

PR #714, opened 2024-03-26
chore: allow release breaking change for stripe node module bump

fermentfan:
This is the PR to solely add the needed commit message for the release process to get the newest Stripe node module in the pipeline. See #705 .

BREAKING CHANGE: this marks the next release to be released with a breaking version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [16, 18, 19]
+        node: [18, 19, 20]
       fail-fast: false
 
     services:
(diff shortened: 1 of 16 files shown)


ericvtheg:
Bump 👀

fermentfan:
@WonderPanda Any idea when you're around to do this? :)

WonderPanda:
> @WonderPanda Any idea when you're around to do this? :)

I updated the commit history for this and a few other commits to make sure that they're following conventional commits and we don't have to add new commits to backfill.

I bumped stripe to 0.7.0 (instead of 0.6.6 which it would have been prior to this change) since we've never done an official 1.0 release. The changelog for the package makes it clear that this is a breaking change and people will have to update their underlying Stripe dependency

Issue #719, opened 2024-04-08
No delivery mode option on AmqpConnection.request method


cs-salim:
`AmqpConnection.request` calls `publish` with a set list of parameters:
https://github.com/golevelup/nestjs/blob/master/packages/rabbitmq/src/amqp/connection.ts#L402
```
await this.publish(
      requestOptions.exchange,
      requestOptions.routingKey,
      payload,
      {
        replyTo: DIRECT_REPLY_QUEUE,
        correlationId,
        headers: requestOptions.headers,
        expiration: requestOptions.expiration,
      }
    );
```

Which doesn't let you pass other request options like delivery mode. Is this intended as a design choice or is it a bug?

underfisk:
@cs-salim It was the bare minimum setup back then, we could call it intentional/design choice but if you need to extend the available options, feel free to fill a PR that receives the partial options instead of manually pass one by one

davidmwhynot:
I opened a PR (#723) to address this as I need to provide to the `userId` option. The PR addresses the use case in this issue as well as my own use case by allowing the entire `amqplib`'s `Options.Publish` interface to be provided, except for the `replyTo` option.

PR #723, opened 2024-04-15
feat(amqp/connection.ts): add publishOptions to RequestOptions; use them in AmqpConnection.request

davidmwhynot:
Added a new property named "publishOptions" to the RequestOptions interface so that the additional
options available in the amqplib's Options.Publish interface can be provided when making rpc
requests. Added said usage of the new property to the AmqpConnection.request method's call to
this.publish.

re #719

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -404,6 +404,7 @@ export class AmqpConnection {
       requestOptions.routingKey,
       payload,
       {
+        ...requestOptions.publishOptions,
         replyTo: DIRECT_REPLY_QUEUE,
         correlationId,
         headers: requestOptions.headers,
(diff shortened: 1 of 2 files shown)


davidmwhynot:
This could also have been done like so:
```ts
export interface RequestOptions extends Omit<Options.Publish, 'replyTo'> {
  exchange: string;
  routingKey: string;
  timeout?: number;
  payload?: any;
}

// usage in AmqpConnection.request:
await this.publish(
  requestOptions.exchange,
  requestOptions.routingKey,
  payload,
  {
    ...requestOptions,
    replyTo: DIRECT_REPLY_QUEUE,
    correlationId,
  }
);
```
however, I did not want to pollute the existing `RequestOptions` interface with the relatively large number of additional options on the `amqplib`'s `Options.Publish` interface.

I'm will to adjust to this style if it's preferred over the current change.

underfisk (approved):
Thank you for this change. It does feel better to me as we don't need to keep mapping 1:1 each time a user needs to override something

WonderPanda:
Published as @golevelup/nestjs-rabbitmq@5.3.0

PR #732, opened 2024-05-03
feat(stripe): extend webhook config for test mode secrets

fermentfan:
add two configuration entries in the webhook config for accountTest and connectTest webhook secrets allowing you to register test mode webhooks to the same endpoint.

re #731

PR #747, opened 2024-06-13
docs(stripe): sync README file with the latest config. Refs #732

matheo:
Implementing the `stripeSecrets` config I saw that the README had references to an old config interface, and also I added a few details to make it easier for a newbie to recognize the keys.


underfisk (approved):
Thank you 🤩

Issue #703, opened 2024-02-27
ts-jest - When function is called with a mock, toHaveBeenCalledWith() always succeed even when it should not


maxime-dupuis:
```
it('mock called with another mock', async () => {
  class Cheese {
    constructor(public name: string) {}
  }
  class Potato {
    private cheese: Cheese;
    public applyCheese(cheese: Cheese) {
      this.cheese = cheese;
    }
  }
  const mockPotato = createMock<Potato>();
  const mockCheese = createMock<Cheese>();

  mockPotato.applyCheese(mockCheese);

  expect(mockPotato.applyCheese).toHaveBeenCalledTimes(1);
  // These all pass, but they should fail because applyCheese wasn't called with these values
  expect(mockPotato.applyCheese).toHaveBeenCalledWith('kaboom');
  expect(mockPotato.applyCheese).toHaveBeenCalledWith(42);
  expect(mockPotato.applyCheese).toHaveBeenCalledWith(null);
  // This one should pass (and does)
  expect(mockPotato.applyCheese).toHaveBeenCalledWith(mockCheese);
});
```

VonRehberg:
Experiencing the same issue. Is there something we can do about it?

VonRehberg:
Hey @underfisk , any plans on when the package will be updated?
The issue is fixed with the merge request #752 
Thanks in advance!

underfisk:
> Hey @underfisk , any plans on when the package will be updated?
> 
> The issue is fixed with the merge request #752 
> 
> Thanks in advance!

Only @WonderPanda can publish

WonderPanda:
Just published. This should be available in @golevelup/ts-jest@0.5.1

VonRehberg:
Thanks a lot man!

PR #752, opened 2024-06-28
fix(ts-sinon): fixed match all issue of jest mocks

VonRehberg:
This PR fixes issue #703

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -8,6 +8,7 @@ interface TestInterface {
   someBool: boolean;
   optional: string | undefined;
   func: (num: number, str: string) => boolean;
+  func2: (entity: TestClass) => void;
 }
 
 class TestClass {
@@ -41,7 +42,7 @@ describe('Mocks', () => {
       const result = mock.switchToHttp().getRequest();
 
       expect(result).toBe(request);
-      expect(mock.switchToHttp).toBeCalledTimes(1);
+      expect(mock.switchToHttp).toHaveBeenCalledTimes(1);
     });
 
     it('should work with truthy values properties', () => {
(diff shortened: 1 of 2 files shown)


underfisk (approved):
lgtm, thank you for this addition :pray:

VonRehberg:
Thanks for maintaining these tools and for merging!

I was considering to mark my commit as a breaking change instead of a patch.
In my opinion it fixes an issue, but it could cause a lot of tests to fail in packages using renovate. Please consider this when publishing to npm

rbnayax:
Hi guys, this causes a major regression. For example, we can't use jest-when lib anymore as the returned mock is now a proxy and not `jest.fn()` and as such we can't set any properties to it.
Please undo this and find a better solution for the issue that initiated this PR.

rbnayax:
The following test passes in 0.5.0, but fails in 0.5.1. Still trying to figure out why it fails.

```ts
import { createMock } from '@golevelup/ts-jest';

describe(`test`, () => {
  it('mock called with another mock', async () => {
    interface Test {
      foo(): number | undefined;
    }

    const mock = createMock<Test>();
    mock.foo.mockImplementation(() => {
      return 6;
    });
    expect(mock.foo()).toEqual(6);
    mock.foo.mockImplementation(() => {
      return 7;
    });
    expect(mock.foo()).toEqual(7);
    mock.foo.mockImplementation(() => {
      return 0;
    });
    expect(mock.foo()).toEqual(0); // Fails
    mock.foo.mockImplementation(() => {
      return undefined;
    });
    expect(mock.foo()).toEqual(undefined); //Fails
  });
});

```

```
Error: Unexpected return from a matcher function.
Matcher functions should return an object in the following format:
  {message?: string | function, pass: boolean}
'{"actual": [Function mockConstructor], "expected": 0, "message": [Function anonymous], "name": "toEqual", "pass": [Function mockConstructor]}' was returned
```

rbnayax:
The new `apply` trap is causing the issue, working on a fix

PR #756, opened 2024-08-07
fix(ts-jest): update cache when setting mock property

### LINKED ISSUE #578: Primitive Property Caching on DeepMocked Objects
I am not sure if this is intended or not, and I did not see an issue already filed regarding this so I am filing it here.

nodejs: v18
@golevelup/ts-jest: 0.3.5

Given the following:
```typescript
class A {
  someProperty: boolean
}

const mockedObjectA: DeepMocked<A> = createMock<A>({ someProperty: true });
```

If I am doing the test and run the following, it would be okay
```typescript
expect(mockedObjectA.someProperty).toBeTrue(); // this is okay
```

If I do this, it seems to be okay too:
```typescript
mockedObjectA.someProperty = false; // assignment before accessing the value seems to be okay
expect(mockedObjectA.someProperty).toBeFalse(); // this is okay
```

If I do this after accessing the property first, then the property is locked into its original value:
```typescript
expect(mockedObjectA.someProperty).toBeTrue(); // assertion okay
mockedObjectA.someProperty = false;
expect(mockedObjectA.someProperty).toBeFalse(); // this fails - instead of false, it is true here
```

Please comment.

cplanson:
When assigning a new value to a mock property, add it also to the cache map. The mock will be able to correctly retrieve the updated value afterward.

fixes #578

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -88,6 +88,16 @@ describe('Mocks', () => {
       expect(mock.func).toHaveBeenCalledWith(42, '42');
     });
 
+    it('should allow mocked properties to be reassigned', () => {
+      const mock = createMock<TestInterface>();
+
+      mock.someNum = 42;
+      expect(mock.someNum).toBe(42);
+
+      mock.someNum = 43;
+      expect(mock.someNum).toBe(43);
+    });
+
     it('should match mocked instances', () => {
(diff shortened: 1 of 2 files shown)


obe-ws (approved):

underfisk (approved):

cplanson:
Hi @underfisk, thanks for the approval ! 🙏 It seems the CI is stuck on this PR, I don't know why. Please let me know if there is anything I can do.

PR #758, opened 2024-08-11
fix(ts-jest): regression in 0.5.1

### LINKED ISSUE #757: ts-jest: regression in 0.5.1
version 0.5.1 introduced a regression that was a result of #752 

Mocked implementation that returns bullish values stopped working.

The following test illustrates the issue:

```ts
import { createMock } from '@golevelup/ts-jest';

describe(`test`, () => {
  it('mock called with another mock', async () => {
    interface Test {
      foo(): number | undefined;
    }

    const mock = createMock<Test>();
    mock.foo.mockImplementation(() => {
      return 6;
    });
    expect(mock.foo()).toEqual(6);
    mock.foo.mockImplementation(() => {
      return 7;
    });
    expect(mock.foo()).toEqual(7);
    mock.foo.mockImplementation(() => {
      return 0;
    });
    expect(mock.foo()).toEqual(0); // Fails
    mock.foo.mockImplementation(() => {
      return undefined;
    });
    expect(mock.foo()).toEqual(undefined); //Fails
  });
});

```

```
Error: Unexpected return from a matcher function.
Matcher functions should return an object in the following format:
  {message?: string | function, pass: boolean}
'{"actual": [Function mockConstructor], "expected": 0, "message": [Function anonymous], "name": "toEqual", "pass": [Function mockConstructor]}' was returned
```

rbnayax:
fix: #757
reverts regression introduced in 0.5.1 and added tests to make sure they don't reoccur

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -88,7 +88,8 @@ describe('Mocks', () => {
       expect(mock.func).toHaveBeenCalledWith(42, '42');
     });
 
-    it('should match mocked instances', () => {
+    // TODO should be enabled once fixed
+    xit('should match mocked instances', () => {
       const mock = createMock<TestInterface>();
       const mockedInstance = createMock<TestClass>({ someProperty: 42 });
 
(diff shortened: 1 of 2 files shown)


rbnayax:
@VonRehberg I had to revert your change in #752. Please find a way to solve your issue that doesn't break the test I added.

VonRehberg:
No need to be desperate ;-) Issue is fixed in #759 
It's not easy to tell if a nullable value is actually a desired nullable value for mocks where you don't know if they exist. But looks like it was solvable in this case :-D

PR #761, opened 2024-08-13
fix(hasura):  handle both "one off" and "cron" type scheduled events

### LINKED ISSUE #760: Hasura Scheduled Event Trigger Payload - does not support Cron Trigger type
When adding some scheduled events I noticed that the nestjs api instance has logged errors like `Handler not found for $UUID`.

Tracing the commit history PR https://github.com/golevelup/nestjs/pull/440 changed the way a scheduled event is detected: https://github.com/golevelup/nestjs/pull/440/files#diff-e4fd4d1361a3c85348251a097ae2b97470d7b10327d06c0b44906093db72f409R42

It seems this PR was made to support "one off scheduled events" which will not have a `name` field, and the author was going to use the `comment` field to know which registered event handler should process the incoming event. The `comment` field is not required, and when it has no value at registration time hasura does not include it in the event payload.
https://hasura.io/docs/latest/api-reference/metadata-api/scheduled-triggers/#metadata-create-scheduled-event

There is also the "cron trigger" type of scheduled event, which is required to have a `name` field and is included in the event payload:
https://hasura.io/docs/latest/api-reference/metadata-api/scheduled-triggers/#metadata-create-cron-trigger

Think this comes down to naming confusion from the hasura docs, since "cron trigger" is the name given in the metadata docs, but the docs for "scheduled trigger" only say there can be "scheduled triggers" and "one off scheduled event"
https://hasura.io/docs/latest/scheduled-triggers/how-it-works/
https://hasura.io/docs/latest/scheduled-triggers/create-one-off-scheduled-event/

---

I think the smallest possible change to support both kinds of scheduled event is to remove `comment` from the `isHasuraScheduledEventPayload` predicate. This is not a required field in either "one off scheduled event" or "schedule event" metadata, and when it is not set with a value it will not be included in the event payload. 

Since `comment` field is not required, I dont think this should break anyone who is already using the library for "one off scheduled events" since they are already registering an event handler with a `name` matching the expected `comment`.

mattgstevens:
…d events

fix #760

### DIFF
--- a/packages/hasura/src/hasura.module.ts
+++ b/packages/hasura/src/hasura.module.ts
@@ -39,7 +39,7 @@ function isHasuraEvent(value: any): value is HasuraEvent {
 function isHasuraScheduledEventPayload(
   value: any
 ): value is HasuraScheduledEventPayload {
-  return ['comment', 'scheduled_time', 'payload'].every((it) => it in value);
+  return ['scheduled_time', 'payload'].every((it) => it in value);
 }
 
 @Module({
@@ -72,7 +72,8 @@ export class HasuraModule
       ],
     }
   )
-  implements OnModuleInit {
+  implements OnModuleInit
+{
   private readonly logger = new Logger(HasuraModule.name);
 
   constructor(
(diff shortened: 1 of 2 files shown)


underfisk (approved):
Lgtm, i would love to hear @WonderPanda feedback on this one e

WonderPanda:
This seems like reasonable behavior to me. Checking the comment always felt a little bit off

mattgstevens:
@WonderPanda is there an ETA on next release? 
not sure what kind of release schedule this package follows

WonderPanda:
> @WonderPanda is there an ETA on next release? not sure what kind of release schedule this package follows

The fix should be available now in `@golevelup/nestjs-hasura@3.0.7`

PR #763, opened 2024-08-14
fix(ts-jest): toString doesnt work on nested properties of a mock

### LINKED ISSUE #762: ts-jest: toString doesnt work on nested properties of a mock
```ts
    it('toString should work', () => {
      const mock = createMock<any>();
      expect(mock.toString()).toEqual('[object Object]');
      expect(mock.nested.toString()).toEqual('function () { [native code] }'); // Fails
    });
```

rbnayax:
fix: #762

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -204,6 +204,12 @@ describe('Mocks', () => {
       expect(third.getClient).toBeDefined();
     });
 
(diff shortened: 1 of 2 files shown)


rbnayax:
@underfisk would appreciate your CR, this issue causes a majority of our integration tests in nestjs to fail
This issue surfaced only now due to the recent changes introduced in 0.5.1

WonderPanda (approved):

PR #766, opened 2024-08-15
fix(ts-jest): setting properties on nested values, doens't override t…

### LINKED ISSUE #765: jest-fn: setting properties on nested values, doens't override the cached values
The following test fails:

```ts
    it('nested properties mocks should be able to set properties and override cache', () => {
      const mock = createMock<any>();
      const autoMockedFn = mock.nested.f;
      expect(typeof autoMockedFn).toEqual('function');
      const myFn = () => 5;
      mock.nested.f = myFn;
      expect(mock.nested.f === myFn).toBeTruthy(); // Fails
    });
```

rbnayax:
…he cached values

fix: #765

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -210,6 +210,15 @@ describe('Mocks', () => {
       expect(mock.nested.toString()).toEqual('function () { [native code] }');
     });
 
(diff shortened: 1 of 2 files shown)


rbnayax:
@WonderPanda @underfisk another fix to a regression presented in 0.5.1 would appreciate your CR

WonderPanda (approved):

PR #776, opened 2024-08-23
Bump node versions and ci matrix

barraponto:
Both node 19 and 21 went [the way of the dodo](https://endoflife.date/nodejs) already.
Let's not hold @golevelup/ts-vitest back because of those versions.
Also, let's extend the ci matrix to node 22, for good measure.


barraponto:
FWIW ci passes in my repo: https://github.com/barraponto/golevelup-nestjs/actions/runs/10532889002?pr=1
(although integration tests seem to blow up, it passes anyway)

package.json
```diff
@@ -107,6 +106,7 @@
   },
   "volta": {
     "node": "18.16.1",
-    "yarn": "1.22.21"
-  }
-}
\ No newline at end of file
+    "yarn": "1.22.22"
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
underfisk: is this necessary?

barraponto: not strictly necessary, but it's a `package.lock` for your project's package manager.
it is defined (and followed) by corepack.

underfisk: corepack is being deprecated, i don't think we require it to be included as part of this PR change

barraponto: i wasn't aware of corepack deprecation (sad).
field removed.

barraponto:
Something needs to be configured because this PR removes node 19 support, but github actions is still expecting it...

underfisk:
> Something needs to be configured because this PR removes node 19 support, but github actions is still expecting it...

Yes, I don't have permissions to bypass this, only  @WonderPanda

underfisk:
Bump @WonderPanda

WonderPanda:
Sorry guys. I've removed the branch protection requirement for Node 19

PR #781, opened 2024-09-10
feat(rabbitmq): adds a message batching mechanism for RabbitMQ handlers

ckfngod:
### Message Batching

This implements consumer-side batching (like [Spring](https://docs.spring.io/spring-cloud-stream/reference/rabbit/rabbit_overview/receiving-batch.html#consumer-side-batching)). It works by accumulating messages until either it hits the batch size limit or the batch timer expires at which time the handler will be presented with the messages as a single array which gets acked or nacked together. A batch error handler may be provided for users that require their error handling logic to be aware of the failed batch as a whole.

This is implemented as a new optional message handler options property `batchOptions` which contains:
- `size` representing the maximum batch length before returning the batch
- `timeout` representing the maximum length of time allowed between messages before returning a batch and
- `errorHandler` a custom error handling implementation that receives message batches

At a high-level the batching mechanism works as follows:
1. Initial message is received:
a. Batch timer is started with message handling logic as the callback
b. Store above callback in a separate variable `inflightBatchHandler`
2. Subsequent message is received, check if the batch size has been reached:
a. If it has, clear the batch timer and immediately call `inflightBatchHandler` with the batch
b. Otherwise, refresh the batch timer
3. No message received for batch timeout duration and timer has not been cleared:
a. `inflightBatchHandler` is called by the timer with a partial batch

### DIFF
--- a/integration/rabbitmq/e2e/subscribe.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/subscribe.e2e-spec.ts
@@ -401,6 +401,8 @@ describe('Rabbit Subscribe', () => {
       }
     };
 
+    const paddedBatchTimeout = BATCH_TIMEOUT + 10;
+
     it('should return a full message batch immediately', async () => {
       const testMessages = await publishMessages(
         BATCH_SIZE,
@@ -415,7 +417,7 @@ describe('Rabbit Subscribe', () => {
     it('should return a partial message batch after timeout', async () => {
       const testMessages = await publishMessages(1, exchange, batchRoutingKey);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchHandler).toHaveBeenCalledTimes(1);
       expect(batchHandler).toHaveBeenCalledWith(testMessages);
     });
@@ -443,7 +445,7 @@ describe('Rabbit Subscribe', () => {
         );
       }
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchHandler).toHaveBeenCalledTimes(3);
       expect(batchHandler).toHaveBeenLastCalledWith(testMessageBatches[2]);
     });
@@ -477,7 +479,7 @@ describe('Rabbit Subscribe', () => {
         batchErrorRoutingKey,
       );
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
       expect(parseMessages(batchErrorHandler.mock.calls[0][1])).toEqual(
         testMessages,
@@ -512,7 +514,7 @@ describe('Rabbit Subscribe', () => {
         );
       }
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchErrorHandler).toHaveBeenCalledTimes(3);
       expect(parseMessages(batchErrorHandler.mock.calls[2][1])).toEqual(
         testMessageBatches[2],
@@ -523,27 +525,27 @@ describe('Rabbit Subscribe', () => {
       await publishMessages(BATCH_SIZE, exchange, batchRoutingKey);
       expect(batchHandler).toHaveBeenCalledTimes(1);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchHandler).toHaveBeenCalledTimes(1);
     });
 
     it('should not return a partial batch before batch timeout', async () => {
       await publishMessages(1, exchange, batchRoutingKey);
 
-      await setTimeout(BATCH_TIMEOUT * 0.9);
+      await setTimeout(paddedBatchTimeout * 0.9);
       expect(batchHandler).toHaveBeenCalledTimes(0);
 
-      await setTimeout(BATCH_TIMEOUT * 0.2);
+      await setTimeout(paddedBatchTimeout * 0.2);
       expect(batchHandler).toHaveBeenCalledTimes(1);
     });
 
     it('should not return another partial batch after first batch timeout', async () => {
       await publishMessages(1, exchange, batchRoutingKey);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchHandler).toHaveBeenCalledTimes(1);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchHandler).toHaveBeenCalledTimes(1);
     });
 
@@ -557,7 +559,7 @@ describe('Rabbit Subscribe', () => {
       await setTimeout(1);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
     });
 
@@ -567,10 +569,10 @@ describe('Rabbit Subscribe', () => {
 
       await publishMessages(1, exchange, batchErrorRoutingKey);
 
-      await setTimeout(BATCH_TIMEOUT * 0.9);
+      await setTimeout(paddedBatchTimeout * 0.9);
       expect(batchErrorHandler).toHaveBeenCalledTimes(0);
 
-      await setTimeout(BATCH_TIMEOUT * 0.2);
+      await setTimeout(paddedBatchTimeout * 0.2);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
     });
 
@@ -580,10 +582,10 @@ describe('Rabbit Subscribe', () => {
 
       await publishMessages(1, exchange, batchErrorRoutingKey);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
 
-      await setTimeout(BATCH_TIMEOUT);
+      await setTimeout(paddedBatchTimeout);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
     });
   });
--- a/integration/rabbitmq/e2e/subscribe.e2e-spec.ts
+++ b/integration/rabbitmq/e2e/subscribe.e2e-spec.ts
@@ -378,11 +378,7 @@ describe('Rabbit Subscribe', () => {
   });
 
   describe('Message Batching', () => {
-    const publishMessages = async (
-      size = BATCH_SIZE,
-      ex = exchange,
-      rk = batchRoutingKey,
-    ) => {
+    const publishMessages = async (size: number, ex: string, rk: string) => {
       const messages: string[] = [];
       for (let i = 0; i < size; i++) {
         const testMessage = `testMessage${i}`;
@@ -403,17 +399,18 @@ describe('Rabbit Subscribe', () => {
     };
 
     it('should return a full message batch immediately', async () => {
-      const testMessages = await publishMessages();
+      const testMessages = await publishMessages(
+        BATCH_SIZE,
+        exchange,
+        batchRoutingKey,
+      );
+
       expect(batchHandler).toHaveBeenCalledTimes(1);
       expect(batchHandler).toHaveBeenCalledWith(testMessages);
     });
 
     it('should return a partial message batch after timeout', async () => {
-      const testMessages = await publishMessages(1);
-
-      // should not have been called at this point due to batch timeout
-      await sleep(BATCH_TIMEOUT / 10);
-      expect(batchHandler).toHaveBeenCalledTimes(0);
+      const testMessages = await publishMessages(1, exchange, batchRoutingKey);
 
       await sleep(BATCH_TIMEOUT);
       expect(batchHandler).toHaveBeenCalledTimes(1);
@@ -424,7 +421,9 @@ describe('Rabbit Subscribe', () => {
       const testMessageBatches: string[][] = [];
 
       for (const size of [BATCH_SIZE, BATCH_SIZE, 1]) {
-        testMessageBatches.push(await publishMessages(size));
+        testMessageBatches.push(
+          await publishMessages(size, exchange, batchRoutingKey),
+        );
       }
 
       // two full batches should be immediately handled
@@ -435,10 +434,6 @@ describe('Rabbit Subscribe', () => {
           testMessageBatches[index],
         );
       }
-
-      // should still only have been called twice at this point due to batch timeout
-      await sleep(BATCH_TIMEOUT / 10);
-      expect(batchHandler).toHaveBeenCalledTimes(2);
 
       await sleep(BATCH_TIMEOUT);
       expect(batchHandler).toHaveBeenCalledTimes(3);
@@ -474,10 +469,6 @@ describe('Rabbit Subscribe', () => {
         batchErrorRoutingKey,
       );
 
-      // should not have been called at this point due to batch timeout
-      await sleep(BATCH_TIMEOUT / 10);
-      expect(batchHandler).toHaveBeenCalledTimes(0);
-
       await sleep(BATCH_TIMEOUT);
       expect(batchErrorHandler).toHaveBeenCalledTimes(1);
       expect(parseMessages(batchErrorHandler.mock.calls[0][1])).toEqual(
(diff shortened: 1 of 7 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -520,6 +567,112 @@ export class AmqpConnection {
...
+            for (const msg of batchMsgsToProcess) {
+              await errorHandler(channel, msg, e);
+            }
+          }
+        };
+
+        batchMsgs.push(msg);
+
+        if (batchMsgs.length === 1) {
+          console.log('received first message, starting timer');
```
underfisk: let's not use console log but rather Nestjs Logger

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -823,12 +1011,20 @@ export class AmqpConnection {
...
+        consumer.channel
+      );
+    } else if (consumer.type === 'subscribe-batch') {
+      newConsumerTag = await this.setupBatchSubscriberChannel<T>(
         consumer.handler,
         consumer.msgOptions,
         consumer.channel
       );
+    } else {
+      throw new Error('Unable to resume consumer, unexpected consumer type.');
```
underfisk: I expect more contextual data as part of this error/exception otherwise the developer will not be able to quickly find the issue

underfisk (changes requested):
I do like this idea of batching, left some notes but also I'll request test coverage as we need to be certain that this works as expected and is backwards compatible

ckfngod:
@underfisk Thanks for taking a look, will do on the logging and tests. Are you able to address some of the other questions that I had?

underfisk:
> Thoughts on the approach?
> What is the preferred way to model this? As a different decorator (something like @BatchRabbitSubscribe) or as part of the existing decorator with some interface changes?
> Is there a preference on error handling behaviour for batches? Should batches be treated as a single unit or individual messages for that purpose?
> I’m not familiar with RabbitMQ RPC but is this something that could be useful there as well? Or is it best to keep this separate as subscriber-only behaviour?

I haven't been using this for a while but I'm familiar with the concept. Spring has a really good implementation and I do believe we'll just benefit.
The API can be improved along usage as I'm not sure how many want/will use batching but even so, the implementation behind doesn't see that hard and we can include that.

1. If a batching process fails, it will be treated as a single failure but ideally should contain in-depth information or some logical exception that makes it easier to debug and trace. I often see batching/transactions as all or nothing unless there's a mode to ignore failures (in case we don't care) and keep going on.
2. Leave the RPC to the next phase, initially this seems just fine what you're bringing to the table (I do appreciate it a lot since I don't have that much time to be an active IC). If you want to collect wider feedback, you can open a discussion in the GH discussions tab and present some follow-up ideas
3. Regarding the decorator, let's add a separate decorator to streamline this new feature. Having a special token for batched subscribers is better than reading if `batchOptions` exist or not. Introducing another optional property is just going to make it more confusing IMO, i'd rather have them separate as they serve different use cases than mixed together

I tried to answer your questions @ckfngod but the "real answer" may come from real use cases or anyone interested in this feature. In theory there is a common ground for designing new API's and I don't think this proposal is far from a good implementation, it is a **good starting point** for sure

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -116,6 +116,11 @@ export interface MessageHandlerOptions {
    * If set, will override the module's default deserializer.
    */
   deserializer?: MessageDeserializer;
+
+  /**
+   * Enables consumer-side batching.
+   */
+  batchOptions?: BatchOptions;
```
underfisk, line 124: For example, if you introduce a separate decorator where its tied to batching, this is no longer required here and its streamlined into the batching.

Now a decorator can have its own Symbol which will help you during the setup phase.
It may be ideal as we're always introducing optional properties

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -460,6 +481,32 @@ export class AmqpConnection {
     originalHandlerName = 'unknown',
     consumeOptions?: ConsumeOptions
   ): Promise<ConsumerTag> {
+    if (msgOptions.batchOptions) {
```
underfisk: Based on my point, if you had a special symbol you wouldn't need to conditionally register as the Decorator would streamline that for you.

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -75,6 +87,15 @@ export type ConsumerHandler<T, U> =
         headers?: any
       ) => Promise<SubscribeResponse>;
     })
+  | (BaseConsumerHandler & {
```
underfisk, line 86: The same would be applied here, less type pollution and a more streamlined type.

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -520,6 +567,112 @@ export class AmqpConnection {
...
 
+  private async setupBatchSubscriberChannel<T>(
+    handler: BatchSubscriberHandler<T>,
+    msgOptions: MessageHandlerOptions,
+    channel: ConfirmChannel,
+    originalHandlerName = 'unknown',
+    consumeOptions?: ConsumeOptions
+  ): Promise<ConsumerTag> {
+    const queue = await this.setupQueue(msgOptions, channel);
+    const batchSize = msgOptions.batchOptions?.size ?? 10;
```
underfisk: Please update the JSDOC to include the defaults `10` for batchSize and `200` for timeout as we don't want to hide them from public API.

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -45,11 +45,23 @@ const DIRECT_REPLY_QUEUE = 'amq.rabbitmq.reply-to';
 export type ConsumerTag = string;
 
 export type SubscriberHandler<T = unknown> = (
+  msg: (T | undefined) | (T | undefined)[],
```
underfisk: We could improve the msg type to something like
```typescript
msg: Optional<T> | Array<Optional<T>>
```

That would read better IMO :sweat_smile:

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -520,6 +567,112 @@ export class AmqpConnection {
...
+            if (response instanceof Nack) {
+              for (const msg of batchMsgsToProcess) {
+                channel.nack(msg, false, response.requeue);
+              }
+              return;
+            }
+
+            // developers should be responsible to avoid subscribers that return therefore
+            // the request will be acknowledged
+            if (response) {
```
underfisk: @ckfngod Do you think this is worth an exception to provide developer awareness? If this shouldn't be happening, a warning itself can slip very easily

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -520,6 +567,112 @@ export class AmqpConnection {
...
+            };
+            batchTimer = setTimeout(batchHandler, batchTimeout);
+            inflightBatchHandler = batchHandler;
+          });
+        } else if (batchMsgs.length === batchSize) {
+          console.log('processing full batch');
+          clearTimeout(batchTimer);
+          await inflightBatchHandler();
+        } else {
+          console.log('not yet full batch, refreshing timer');
```
underfisk: Sorry for poking but these logs should probably be under `logger.debug` and not the normal info as they're purely useful for debugging/tracing

ckfngod:
@underfisk Really appreciate the in-depth answers, that's very helpful.

1. Agreed, I think it makes alot of sense conceptually to keep it as a single unit in general. Also makes it a little easier to not have to worry about individual message IDs.
2. Sounds good, I'll leave RPC alone for now.
3. Makes sense to me. I was leaning in this direction as well because of type complications with `errorHandler` in the external interface.

I'll go ahead with the proposed changes. Let me know if there's any other changes you'd like to see!

> I tried to answer your questions @ckfngod but the "real answer" may come from real use cases or anyone interested in this feature.

Totally agree with this sentiment. For what it's worth, I'm working on this because we have a use case for it and this solution will work fine for us 🙂

ckfngod:
> Regarding the decorator, let's add a separate decorator to streamline this new feature. Having a special token for batched subscribers is better than reading if batchOptions exist or not. Introducing another optional property is just going to make it more confusing IMO, i'd rather have them separate as they serve different use cases than mixed together

@underfisk Spent some time thinking about this and this might not work as cleanly as we thought. The types get pretty hairy and will impact external interfaces (would need a discriminated union on `RabbitHandlerConfig` for example) which may be a little risky. It also gets a little ambiguous when considering handler configs defined at the module level because of how configs are merged (if I use `@RabbitSubscribe` but then later define `batchOptions` at the module-level, is it batched or not? Do we impose some restriction at runtime based on the handler type?). The ability to enable batching at the module-level would be a very nice to have for us.

What I like about the current solution is that the types are simple (only one handler options interface, no complex discriminated unions), it’ll work for both decorator and module-level configs unambiguously (if `batchOptions` is present, we batch) and I actually like how it’s built into the existing decorator as it’ll be very easy for interested users to enable the behaviour.

Thoughts?

underfisk:
> > Regarding the decorator, let's add a separate decorator to streamline this new feature. Having a special token for batched subscribers is better than reading if batchOptions exist or not. Introducing another optional property is just going to make it more confusing IMO, i'd rather have them separate as they serve different use cases than mixed together
> 
> @underfisk Spent some time thinking about this and this might not work as cleanly as we thought. The types get pretty hairy and will impact external interfaces (would need a discriminated union on `RabbitHandlerConfig` for example) which may be a little risky. It also gets a little ambiguous when considering handler configs defined at the module level because of how configs are merged (if I use `@RabbitSubscribe` but then later define `batchOptions` at the module-level, is it batched or not? Do we impose some restriction at runtime based on the handler type?). The ability to enable batching at the module-level would be a very nice to have for us.
> 
> What I like about the current solution is that the types are simple (only one handler options interface, no complex discriminated unions), it’ll work for both decorator and module-level configs unambiguously (if `batchOptions` is present, we batch) and I actually like how it’s built into the existing decorator as it’ll be very easy for interested users to enable the behaviour.
> 
> Thoughts?

Sounds fair to me. You're presenting the idea with an use case in mind, I think we can refine it later if people start adhering to it

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -422,17 +437,43 @@ export class AmqpConnection {
...
+        channel,
+        originalHandlerName,
+        consumeOptions
+      )
+    );
+  }
+
+  private async setupChannel(
+    msgOptions: MessageHandlerOptions,
+    setupFunction: (channel: ConfirmChannel) => Promise<string>
```
underfisk, line 462: let's rename this parameter to something more readable (e.g: `consumerFactory` not sure but there may be a better indicative)

integration/rabbitmq/e2e/subscribe.e2e-spec.ts
```diff
@@ -35,6 +35,18 @@ const deleteHandler = jest.fn();
...
+const BATCH_TIMEOUT = 200;
+const batchHandler = jest.fn();
+const batchRoutingKey = 'testSubscribeBatch';
+const batchQueue = 'testSubscribeBatchQueue';
+const batchErrorHandler = jest.fn();
+const batchErrorRoutingKey = 'testSubscribeBatchError';
+const batchErrorQueue = 'testSubscribeBatchErrorQueue';
+
+const sleep = async (ms: number) =>
+  new Promise((resolve) => setTimeout(resolve, ms));
```
underfisk: @ckfngod You can use `node:timers/promises` and the `setTimeout` will act as the sleep function, Node has that now built in :)

example:

```typescript
import { setTimeout } from 'node:timers/promises'

await setTimeout(ms you want)
```

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -195,3 +201,26 @@ export interface RabbitMQChannelConfig {
...
+   * This should be smaller than the channel prefetch.
+   *
+   * Defaults to 10 if provided value is less than 2.
+   */
+  size: number;
+
+  /**
+   * The time to wait, in milliseconds, for additional messages before returning a partial batch.
+   *
+   * Defaults to 200 if not provided or provided value is less than 1.
```
underfisk, line 226: use jsdoc `@default`

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -195,3 +201,26 @@ export interface RabbitMQChannelConfig {
...
+
+interface BatchOptions {
+  /**
+   * The number of messages to accumulate before calling the message handler.
+   *
+   * This should be smaller than the channel prefetch.
+   *
+   * Defaults to 10 if provided value is less than 2.
+   */
+  size: number;
```
underfisk, line 221: use jsdoc `@default`

underfisk:
@ckfngod Before we get this in, we'll need to polish the PR description as there are open questions possibly resolved and we're going to present a new feature. If you're okay later on to improve the documentation (md files) with some examples, that would be awesome

packages/rabbitmq/src/amqp/connection.ts
ckfngod: yeah these types look a little redundant just in general. i'll see what i can do here

ckfngod:
@underfisk Added tests, documentation and updated PR description with more detail / summaries of our discussion. Let me know if there are any other tasks or outstanding issues you'd like to see resolved!

packages/rabbitmq/src/amqp/connection.ts
ckfngod: simplified them a bit and tied them to the main handler types. there might be an opportunity to go even further with conditional types but i don't think it's really necessary

packages/rabbitmq/src/amqp/errorBehaviors.ts
```diff
@@ -69,17 +90,18 @@ export const defaultAssertQueueErrorHandler: AssertQueueErrorHandler = (
...
-};
+export const forceDeleteAssertQueueErrorHandler: AssertQueueErrorHandler =
+  async (
+    channel: Channel,
+    queueName: string,
+    queueOptions: QueueOptions | undefined,
+    error: any
+  ) => {
+    if (error.code == 406) {
+      //406 == preconditions failed
```
underfisk: Can we extract `406` to a const (e.g: `PRECONDITION_FAILED_CODE`)

packages/rabbitmq/src/rabbitmq.interfaces.ts
underfisk: @ckfngod JSDoc is not yet implemented using `@default`

packages/rabbitmq/src/rabbitmq.interfaces.ts
underfisk: @ckfngod JSDoc is not yet implemented using `@default`

underfisk (changes requested):

packages/rabbitmq/src/amqp/connection.ts
ckfngod: This was copied from the existing `setupSubscriberChannel` function. I can remove it if you think it doesn't provide any value.

packages/rabbitmq/src/amqp/connection.ts
ckfngod: I've removed these guys 👍

packages/rabbitmq/src/amqp/connection.ts
ckfngod: I've removed this type.

packages/rabbitmq/src/amqp/connection.ts
ckfngod: I removed it

integration/rabbitmq/e2e/subscribe.e2e-spec.ts
ckfngod: Good to know!

packages/rabbitmq/src/amqp/errorBehaviors.ts
```diff
@@ -69,17 +92,18 @@ export const defaultAssertQueueErrorHandler: AssertQueueErrorHandler = (
...
-};
+export const forceDeleteAssertQueueErrorHandler: AssertQueueErrorHandler =
+  async (
+    channel: Channel,
+    queueName: string,
+    queueOptions: QueueOptions | undefined,
+    error: any
+  ) => {
+    if (error.code == PRECONDITION_FAILED_CODE) {
+      //406 == preconditions failed
```
underfisk: You can now remove this comment since the code is extracted and easy to read

underfisk (approved):

underfisk:
@ckfngod Thank you for addressing the PR feedback :pray: 
Once all tests pass I'll have it merged

ckfngod:
@underfisk Cool, appreciate it!

ckfngod:
@underfisk pushed a change that should fix my flaky tests. was working locally but looks like there's a timing issue in ci pipeline. see https://github.com/golevelup/nestjs/pull/781/commits/503c827d741a9add82e3549e40ad07ccbb531ce5

david-pivonka:
It looks like the logic for this doesn't work with the default RabbitMQ that this module creates.

This is the RabbitMQ module we use:
```
@Module({
  imports: [
    HttpModule,
    RabbitMQModule.forRoot({
      uri: ConfigurationService.getString('RABBITMQ_URI'),
      exchanges: [
        {
          name: 'amq.direct',
          type: 'direct',
        },
      ],
      connectionInitOptions: {
        wait: false,
      },
    }),
  ],
  providers: [PublisherService, QueueService],
  exports: [PublisherService, QueueService],
})
export class QueueModule {
}
```

This is the batch consumer:
```
import { RabbitSubscribe } from '@golevelup/nestjs-rabbitmq';
import { Injectable } from '@nestjs/common';

@Injectable()
export class TestConsumer {
  @RabbitSubscribe({
    exchange: 'amq.direct',
    routingKey: 'batch-route',
    queue: 'batch-queue',
    batchOptions: {
      size: 156,
    },
  })
  batchErrorSubscriber(messages) {
    console.log(`Received ${messages.length} messages`);
  }
}
```

Even though I specified the batch size to be 156, I still receive a maximum of 100 messages. I can see that the queue contains more than 10 messages. This number (10) is directly affected by the default `prefetchCount` in this file:
IntelliJ IDEA 2025-02-06 13 55 49

Can anyone of you please help me resolve this issue? It all seems that the issue is related to the channel that is created, which limits it to 10.

underfisk:
@david-pivonka If we were to remove the `prefetchCount` default value, would that help your use-case?

I think it was added as part of a feature implementation but I'm totally fine by removing the default and letting the 1/2 user(s) that requested/implemented this know in a breaking change that we no longer inject it by default.

david-pivonka:
@underfisk Hey, thanks for the comment. I've just set the `prefetchCount` to `undefined`, and everything is working as expected. So yes, I would appreciate it if you could remove the default value as it seems like breaking for the batch logic.

Let me know if you want me to create a PR for it.

This is the working code:
```
import { RabbitMQModule } from '@golevelup/nestjs-rabbitmq';
import { HttpModule } from '@nestjs/axios';
import { Module } from '@nestjs/common';
import ConfigurationService from '@rennie/configuration/service/configuration.service';
import { QueueService } from '@rennie/queue/service/queue.service';
import { PublisherService } from './service/publisher.service';

@Module({
  imports: [
    HttpModule,
    RabbitMQModule.forRoot({
      uri: ConfigurationService.getString('RABBITMQ_URI'),
      exchanges: [
        {
          name: 'amq.direct',
          type: 'direct',
        },
      ],
      prefetchCount: undefined,
      connectionInitOptions: {
        wait: false,
      },
    }),
  ],
  providers: [PublisherService, QueueService],
  exports: [PublisherService, QueueService],
})
export class QueueModule {
}
```

underfisk:
> @underfisk Hey, thanks for the comment. I've just set the `prefetchCount` to `undefined`, and everything is working as expected. So yes, I would appreciate it if you could remove the default value as it seems like breaking for the batch logic.
> 
> 
> 
> Let me know if you want me to create a PR for it.
> 
> 
> 
> This is the working code:
> 
> ```
> 
> import { RabbitMQModule } from '@golevelup/nestjs-rabbitmq';
> 
> import { HttpModule } from '@nestjs/axios';
> 
> import { Module } from '@nestjs/common';
> 
> import ConfigurationService from '@rennie/configuration/service/configuration.service';
> 
> import { QueueService } from '@rennie/queue/service/queue.service';
> 
> import { PublisherService } from './service/publisher.service';
> 
> 
> 
> @Module({
> 
>   imports: [
> 
>     HttpModule,
> 
>     RabbitMQModule.forRoot({
> 
>       uri: ConfigurationService.getString('RABBITMQ_URI'),
> 
>       exchanges: [
> 
>         {
> 
>           name: 'amq.direct',
> 
>           type: 'direct',
> 
>         },
> 
>       ],
> 
>       prefetchCount: undefined,
> 
>       connectionInitOptions: {
> 
>         wait: false,
> 
>       },
> 
>     }),
> 
>   ],
> 
>   providers: [PublisherService, QueueService],
> 
>   exports: [PublisherService, QueueService],
> 
> })
> 
> export class QueueModule {
> 
> }
> 
> ```

Feel free to create the PR

PR #793, opened 2024-09-30
chore(ci): github inactive issues auto-closing

underfisk:
Integrates `actions/stale` in CI to help closing issues with no activity.

### DIFF
--- a/.github/workflows/inactive-issues.yml
+++ b/.github/workflows/inactive-issues.yml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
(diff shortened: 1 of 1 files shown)


.github/workflows/inactive-issues.yml
```diff
@@ -0,0 +1,22 @@
...
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
underfisk, line 22: @WonderPanda This will require a GITHUB_TOKEN to be defined i assume :thinking: 
The value of stale/close can be adjusted, these are just the default which sounded reasonable

WonderPanda: It looks like it should be provided automatically: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

We can give this a try and see if it works without needing to tweak permissions or anything

WonderPanda (approved):

PR #795, opened 2024-09-30
refactor(rmq): logging enhancements and connection manager as Map over Array

underfisk:
### Improvements
- Connection manager utilizes a `Map` for faster lookups `O(1)` over an array. This can be useful for users with a large amount of connections
- Missing logs were added to provide contextual information. Others were improved with more details
- Removed deprecated `RabbitMQModule.build` to avoid confusion and adopt the right pattern 

### Breaking changes
- `RabbitMQModule.build` was removed and `forRoot` or `forRootSync` should be used

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -562,11 +562,18 @@ export class AmqpConnection {
     let batchTimer: NodeJS.Timeout;
     let inflightBatchHandler: () => Promise<void>;
 
+    // Normalize batch values but warn consumer about this adjusts
     if (!batchSize || batchSize < 2) {
+      this.logger.warn(
+        `batch size too low/not defined, received: ${batchSize}. Adjusting to 10`
+      );
       batchSize = 10;
     }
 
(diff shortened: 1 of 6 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -562,11 +562,18 @@ export class AmqpConnection {
     let batchTimer: NodeJS.Timeout;
     let inflightBatchHandler: () => Promise<void>;
 
+    // Normalize batch values but warn consumer about this adjusts
```
underfisk, line 565: The logs below are meant to provide further feedback to developers, they should know when we normalize these values for the ease of debugging

packages/rabbitmq/src/rabbitmq.module.ts
```diff
@@ -100,28 +96,6 @@ export class RabbitMQModule
     return connection;
   }
 
-  public static build(config: RabbitMQConfig): DynamicModule {
```
underfisk, line 103: It was advised years ago to adopt `forRoot`  which is the common nestjs way therefore this is expected if anyone is using it

PR #797, opened 2024-09-30
chore(deps): upgrade outdated dependencies

underfisk:
### Improvements
- Contains tons of upgrades to dev/peer/optional dependencies as they have not been updated for quite some time


package.json
```diff
@@ -81,31 +80,14 @@
...
-  },
   "version": "0.0.0",
   "config": {
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
   "volta": {
-    "node": "18.16.1",
+    "node": "20.17.0",
```
underfisk, line 90: Starting to set the default to `v20`

vite.config.js
underfisk, line 1: i would like to set vitest as the default test runner and deprecate Jest

PR #798, opened 2024-09-30
chore(ci): drop nodejs 18 support

underfisk:
Nodejs LTS is now set to `v20.x.x`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  LATEST_NODE_VERSION: 18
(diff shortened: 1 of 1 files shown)


underfisk:
@WonderPanda You'll have to also remove the CI nodejs18 protection on this PR :sweat_smile: 
I don't think we should keep non-lts support and it's a good practise to guide our users to only use secure node versions that are also maintained

WonderPanda (approved):

WonderPanda:
> @WonderPanda You'll have to also remove the CI nodejs18 protection on this PR 😅 I don't think we should keep non-lts support and it's a good practise to guide our users to only use secure node versions that are also maintained

Done!

PR #850, opened 2024-10-08
fix(ts-jest): ensures auto mocked properties can be casted to primitive types

### LINKED ISSUE #843: ts-jest 0.5.1 regression
I noticed a changed of behaviour when upgrading `@golevelup/ts-jest` from 0.5.0 to 0.5.1

Given the following piece of code 
```js 
import { createMock } from './mocks';

interface TestInterface {
  func: () => Promise<Counter>;
}

type Counter = {
  count: number;
};

describe('regression', () => {
  it('showcases the regression', async () => {
    expect.hasAssertions();
    const testmock = createMock<TestInterface>();

    const c = await testmock.func();
    try {
      c.count < 0;
    } catch (e) {
      if (e instanceof Error)
        expect(e.message).toBe('Cannot convert object to primitive value');
    }
  });
});
```

On version 0.5.0 `c.count < 0` doesn't throw.
Screenshot 2024-10-07 at 13 30 53

Whereas on version 0.5.1 `c.count < 0` throws with a TypeError. This behaviours remains with the latest release as well.
Screenshot 2024-10-07 at 13 32 09

Seeing it was introduced through a "patch" bump I wonder if this new behaviour is meant to stay or if it was introduced by mistake ?

JonasFaure:
Addresses a unintended behaviour change introduced in version 0.5.1

fix #843

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -9,6 +9,7 @@ interface TestInterface {
   optional: string | undefined;
   func: (num: number, str: string) => boolean;
   func2: (entity: TestClass) => void;
+  func3: () => Promise<{ prop: number }>;
 }
 
 class TestClass {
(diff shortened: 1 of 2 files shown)


JonasFaure:
I branched out from the tag `@golevelup/ts-jest@0.5.0` and backported those tests there to confirm that they do describe the previous behaviour ✅ 

branch available here : https://github.com/JonasFaure/nestjs/tree/fix/843-proof-behaviour-is-consistent-with-0.5.0
if you want to see for yourselves

Hareloo:
After this update, when doing something like this:
const error = createMock<AxiosError>({ code: null, response: {}, message: 'error' });
I am getting this error: `TypeError: Cannot set properties of null (setting 'Symbol(Symbol.toPrimitive)')`

PR #868, opened 2024-10-29
fix(deps): fix pnpm dependency issue

### LINKED ISSUE #817: Publish @golevelup/nestjs-rabbitmq
Any chance for a rabbitmq release soon so that I can start playing around with message batching 🙂 

@WonderPanda @underfisk

ckfngod:
fixes #817

Never used pnpm before but my understanding is that it's looking for the new version of a local workspace package on the registry which it can't find because it hasn't been published yet.

I believe the correct way to handle this is to update all golevelup dependencies to use the local workspace packages instead of what is on the registry as detailed in their docs (https://pnpm.io/workspaces#publishing-workspace-packages).

@WonderPanda @underfisk Let me know what you guys think.

### DIFF
--- a/packages/graphql-request/package.json
+++ b/packages/graphql-request/package.json
@@ -38,9 +38,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@golevelup/nestjs-common": "^2.0.0",
-    "@golevelup/nestjs-discovery": "^4.0.1",
-    "@golevelup/nestjs-modules": "^0.7.1",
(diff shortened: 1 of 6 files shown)


underfisk:
@ckfngod It looks good to me, i haven't tested but does everything run locally/build?
You'll have to install the dependencies again because CI is failing due to the lockfile not being updated

ckfngod:
@underfisk added lockfile. yeah, everything builds and all tests pass.

Issue #751, opened 2024-06-28
Error when use @RabbitRPC in one controller and listten exchange and queue that same for all actions with diffirent routing keys


sur-ser:
when i use RabbitRPC in one controller with multiple actions like this

```
@RabbitRPC({
    exchange: AuthRegister.exchange,
    routingKey: AuthRegister.topic,
    queue: AuthRegister.queue,
    errorBehavior: MessageHandlerErrorBehavior.NACK,
    errorHandler: rpcErrorHandler,
  })
  async register(@Body() dto: AuthRegister.Request) : Promise<AuthRegister.Response> {
    const newUser = await this.authService.register(dto);
    return { email: newUser.email };
  }


  @RabbitRPC({
    exchange: AuthJWTLogin.exchange,
    routingKey: AuthJWTLogin.topic,
    queue: AuthJWTLogin.queue,
    errorBehavior: MessageHandlerErrorBehavior.NACK,
    errorHandler: rpcErrorHandler,
  })
  async login(@Body() dto: AuthJWTLogin.Request): Promise<AuthJWTLogin.Response> {
    const { id } = await this.authService.validateUser(dto.email, dto.password);
    return this.authService.login(id);
  }
```
  
  and create request like this
  
 ```
 return await this.amqpConnection.request<AuthRegister.Response>({
        exchange: AuthRegister.exchange,
        routingKey: AuthRegister.topic,
        payload: dto,
        timeout: 10000
      })
```
      
      one time it works fine and second time i have got error like this
      

> [Nest] 119672  - 06/27/2024, 7:57:17 PM   ERROR [AmqpConnection] Received message with invalid routing key: sso.auth.register.command

      
but if i keep one action it works fine or when i change que name and keep one routing key in queue it works fine

my exchange is topic
and in controller **exchange and queue is same** for all actions only routing key is diffirent

**Comment by slyk** [comment]
Same for me.
v 4.1 works fine.
v 5 gives errors. I found that the matchesRoutingKey() function is called ant tries to match the message routing key always with different handlers. So it is not taking messages from queue that match given routing keys, it just takes all messages from queue and tries to match them with a random routingKey from a file.

In my project I have three @RabbitRPC() listeners in one file and I sometimes get listener working ok and some times errors with "invalid routing key".

Here is my code and errors as example:
```
@RabbitRPC({
    exchange: 'amq.topic',      //using default exchange for topics
    routingKey: 'user.upsert',  //when message has this routing key (topic)
[...]

PR #869, opened 2024-10-30
Fix/751 rpc with different routing keys

vhforbes:
Hey there,

Regarding the issue: #751

I'm not experienced enough to really find the core problem so that the @RabbitRPC only listen to messages with the correct routingKeys.

Anyhow, I've wanted to propose a change so that if the routingKey is invalid, the message to be requeued so it can be picked up by the correct listener.

Any suggestions on how I can solve this in a better way are welcomed. I would do this fix only In my own code but realized it might be useful for someone else.

Thanks a lot for the lib!

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -702,9 +702,9 @@ export class AmqpConnection {
           if (
             !matchesRoutingKey(msg.fields.routingKey, rpcOptions.routingKey)
           ) {
-            channel.nack(msg, false, false);
+            channel.nack(msg, false, true);
             this.logger.error(
-              'Received message with invalid routing key: ' +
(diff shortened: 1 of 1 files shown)


underfisk:
Thank you @vhforbes for the contribution. In order to us to accept such behavioural change, we would need a test coverage update and possibly one more person's feedback since this will switch the behaviour.

vhforbes:
Hey @underfisk , what do mean by a test coverage update? As I can see right now there are no existing tests for the setupRpcChannel. I don't know how I can provide someone else feedback other than what's already in the issue #751, should I ask there if the change makes sense?

vhforbes:
I've just reread some RMQ docs and believe it its the intended behavior to have separate queues for different processes instead of splitting them using only routing keys. I'll close my PR since it was a misassumption but will be tuned in to improve the lib as I can!

Thanks a lot :)

PR #872, opened 2024-11-05
new AmqpConnection() should NOT cause an error

AckerApple:
When `new AmqpConnection()` with no arguments, it should NOT error but it does

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -147,7 +147,7 @@ export class AmqpConnection {
     this.config = {
       deserializer: (message) => JSON.parse(message.toString()),
       serializer: (value) => Buffer.from(JSON.stringify(value)),
-      logger: config.logger || new Logger(AmqpConnection.name),
+      logger: config?.logger || new Logger(AmqpConnection.name),
       ...defaultConfig,
       ...config,
     };
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -147,7 +147,7 @@ export class AmqpConnection {
     this.config = {
       deserializer: (message) => JSON.parse(message.toString()),
       serializer: (value) => Buffer.from(JSON.stringify(value)),
-      logger: config.logger ?? new Logger(AmqpConnection.name),
+      logger: config?.logger || new Logger(AmqpConnection.name),
       ...defaultConfig,
       ...config,
     };
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -147,7 +147,7 @@ export class AmqpConnection {
     this.config = {
       deserializer: (message) => JSON.parse(message.toString()),
       serializer: (value) => Buffer.from(JSON.stringify(value)),
-      logger: config?.logger || new Logger(AmqpConnection.name),
+      logger: config.logger ?? new Logger(AmqpConnection.name),
       ...defaultConfig,
       ...config,
     };


packages/rabbitmq/src/amqp/connection.ts
underfisk, line 1: @AckerApple There are a lot of formatting diffs, please make sure you're using our prettier/eslint config

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -147,7 +147,7 @@ export class AmqpConnection {
     this.config = {
       deserializer: (message) => JSON.parse(message.toString()),
       serializer: (value) => Buffer.from(JSON.stringify(value)),
-      logger: config.logger || new Logger(AmqpConnection.name),
+      logger: config?.logger || new Logger(AmqpConnection.name),
```
underfisk, line 150: You're saying that the fix is by not conditionally accessing `config`? If so, why doesn't typescript complain here that if config does not exist, logger will eventually crash?

AckerApple: TypeScript fails because the constructor enforces saying that a config argument will be passed in

Simply put, `new AmqpConnection()` appears like it should work but it does not

packages/rabbitmq/src/amqp/connection.ts
AckerApple: I have fixed

packages/rabbitmq/src/amqp/connection.ts
underfisk: @AckerApple So you're saying that the rabbitmq can also be `undefined`? Can we update the constructor param signature to be either the object or undefined by marking it optional explicitly so that way we can control it?
If you want, as part of the constructor you can coalesce to the default instead of having to to these inline's, that is definitely an improvement that I forsee

AckerApple: No I'm sorry there is a typing conflict when the first argument is made optional and since I am not a maintainer of this code, I cannot just guess at the proper solution

More about the typing conflict: Because `this.config` is mapped to the first argument, if you/me/we say that the first argument is optional it then creates a chain reaction of typing conflicts. Try and change it for yourself to see more.

Please continue to help this work along. I appreciate you

AckerApple: Again this support of no config arguments is needed because of the NestJs Injector code seen below which can result in no config arguments:

```js
        if ((0, shared_utils_1.isNil)(inject) && isInContext) {
            // console.log('instances', instances)
            instanceHost.instance = wrapper.forwardRef
                ? Object.assign(instanceHost.instance, new metatype(...instances))
                : new metatype(...instances); // <--------- no arguments aka config is being sent to new AmqpConnection()
        }
```

underfisk: @AckerApple Review my latest change, ideally, all we have to do is use null coalescence operator `??` operator instead of or `||`

AckerApple: I have reviewed your change and that would NOT work because the argument `config` is undefined.

See screen capture where I tried it just for your review:

image

AckerApple: @underfisk I used to be on an old version that worked so I dug up that old code and look, config didn't use to error when this code was there BECAUSE config.logger is NOT being immediately read

image

underfisk: I see, let's then add the conditional check that you're describing where you do `config?.logger ?? new Logger` but add a comment explaining why its needed.
It still feels to me that at some place we're lying to typescript and we're injecting undefined, the constructor makes it look like it's required when in reality it isn't ...

AckerApple: I see you merged. I did think I found the correct typing you were looking for but you might not like the use of `as RabbitMqConfig`

Check it out below, I think this is a more accurate data typing

```ts
  constructor(configArg?: RabbitMQConfig) {
    this.config = {
      deserializer: (message) => JSON.parse(message.toString()),
      serializer: (value) => Buffer.from(JSON.stringify(value)),
      logger: configArg?.logger ?? new Logger(AmqpConnection.name),
      ...defaultConfig,
      ...configArg as RabbitMQConfig,
    };

    this.logger = this.config.logger;
  }
```

underfisk: I'm good for now to simply keep what we've merged and later on, if you'd like, we can propose a refactor that makes this more type-safety and less fragile

AckerApple: Thank you humbly for your attentiveness on this subject. Be well friend

PR #874, opened 2024-11-07
Allow more valid URIs in `assertRabbitMqUri`

### LINKED ISSUE #873: RabibtMQ: New URI validation is breaking amqps uri's
Since the latest version 5.6.0 of the [@golevelup/nestjs-rabbitmq](https://www.npmjs.com/package/@golevelup/nestjs-rabbitmq/v/5.6.0) package, we can no longer connect to our Amazon MQ.
According to the documentation of amazon we need to start our URL with `amqps://` instead of `amqp://`

This has always worked, but the latest version [introduced ](https://github.com/golevelup/nestjs/commit/af3ac52045e6893aa4e82f2a16964d4826641c87#diff-dba83c9f93093f5f8b2d8b1bc2aa654890601a59ad41f5b7f3de460827f632fbR31)a validation of the URL, but is only allowing `amqp` not `amqps`

FloEdelmann:
* Fixes #873.
* Follow-up to #844.

CC @underfisk as author of #844.

### DIFF
--- a/packages/rabbitmq/src/amqp/utils.ts
+++ b/packages/rabbitmq/src/amqp/utils.ts
@@ -29,7 +29,7 @@ export function matchesRoutingKey(
 }
 
 const rabbitMQRegex =
-  /^amqp:\/\/(([^:]+):([^@]+)@)?([^:/]+)(:[0-9]+)?(\/[^\/]+)?$/;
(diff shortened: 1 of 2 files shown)


underfisk (approved):
Thank you!

PR #879, opened 2024-11-10
fix(ts-jest): ts-jest: toPrimitive should not be set on proxies, but …

rbnayax:
…mocked when needed

regression introduced in #850
@JonasFaure you can't just add properties to the mocks as it interferes with libraries that makes comparison, like jest-when

### DIFF
--- a/packages/testing/ts-jest/src/mocks.spec.ts
+++ b/packages/testing/ts-jest/src/mocks.spec.ts
@@ -226,6 +226,15 @@ describe('Mocks', () => {
       expect(mock.toString()).toEqual('[object Object]');
       expect(mock.nested.toString()).toEqual('function () { [native code] }');
     });
+
+    it('mock should equal its partial', () => {
+      const mock = createMock<any>();
+      expect({ mock }).toEqual({ mock: {} });
+      const partialMock = createMock<any>({ foo: { bar: 1 } });
+      expect({ partialMock }).toEqual({ partialMock: { foo: { bar: 1 } } });
+      expect({ foo: partialMock.foo }).toEqual({ foo: { bar: 1 } });
+    });
+
     it('nested properties can be implictly casted to string', () => {
       const mock = createMock<{ nested: any }>();
 
@@ -377,7 +386,7 @@ describe('Mocks', () => {
 
       mockedProvider = module.get<DeepMocked<ExecutionContext>>(diToken);
       dependentProvider = module.get<{ dependent: () => string }>(
-        dependentToken
+        dependentToken,
       );
     });
 
@@ -389,7 +398,7 @@ describe('Mocks', () => {
       mockedProvider.switchToHttp.mockReturnValueOnce(
         createMock<HttpArgumentsHost>({
           getRequest: () => request,
-        })
+        }),
       );
 
       const mockResult = mockedProvider.switchToHttp().getRequest();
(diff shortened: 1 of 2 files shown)


rbnayax:
@underfisk can you please review as current version breaks behavior

rbnayax:
I will say that the behavior introduced is not relevant to this library in my opinion. If you have mocks that needs to be able to cast to primitive, you need to put that symbol in the partial of the proxy you create.
@underfisk @JonasFaure

packages/testing/ts-jest/src/mocks.ts
```diff
@@ -46,6 +46,16 @@ const jestFnProps = new Set([
   'calls',
 ]);
 
+const toPrimitive = (hint) => {
```
underfisk: Can we add the type of hint explicitly?

underfisk (commented):
Can we add test coverage to be certain of what we are doing here in order to avoid future regression?

packages/testing/ts-jest/src/mocks.ts
rbnayax: done

rbnayax:
> Can we add test coverage to be certain of what we are doing here in order to avoid future regression?

done

packages/testing/ts-jest/src/mocks.spec.ts
```diff
@@ -226,6 +226,13 @@ describe('Mocks', () => {
       expect(mock.toString()).toEqual('[object Object]');
       expect(mock.nested.toString()).toEqual('function () { [native code] }');
     });
+
+    it('nested properties should equal its partial', () => {
+      const mock = createMock<any>({ foo: { bar: 1 } });
```
underfisk, line 231: @rbnayax Is this enough to prove that primitives are built as expected?
Does this test run if you don't add your change, i'd like to make sure it does enough

rbnayax: I verified that it failed on the previous code.  
The previous PR introduced new tests to verify the primitive behaviour and they are passing after my change so I believe that is enough

VonRehberg:
Thanks @rbnayax for tackling this!

rbnayax:
I still strongly believe we should remove this feature altogether.
If you create a mock and don't set a property on it, then it should not become a number all of a sudden. This can cause tests to pass when they shouldn't. If you have tests failing because of the casting, set a proper value for them.
Thoughts @underfisk @JonasFaure ?

underfisk:
> I still strongly believe we should remove this feature altogether. If you create a mock and don't set a property on it, then it should not become a number all of a sudden. This can cause tests to pass when they shouldn't. If you have tests failing because of the casting, set a proper value for them. Thoughts @underfisk @JonasFaure ?

I agree, this feels too magical in a way. Setting to `0` in a number is not the same as undefined/null aka nil which may mislead when you're asserting something.

rbnayax:
> > I still strongly believe we should remove this feature altogether. If you create a mock and don't set a property on it, then it should not become a number all of a sudden. This can cause tests to pass when they shouldn't. If you have tests failing because of the casting, set a proper value for them. Thoughts @underfisk @JonasFaure ?
> 
> I agree, this feels too magical in a way. Setting to `0` in a number is not the same as undefined/null aka nil which may mislead when you're asserting something.

Done, ready for review

rbnayax:
@WonderPanda Can you please release a new version so the regression introduced in the previous version will be fixed?

PR #920, opened 2025-01-05
refactor: migrate hasura module to use nestjs `ConfigurableModuleClass`

underfisk:
## Improvements
- Hasura Module utilizes `ConfigurableModuleClass` from Nestjs

## Breaking Changes
- Initializing `HasuraModule` no longer requires the module to be provided

### New API
```typescript
HasuraModule.forRoot({
  your config
})
```


underfisk:
@WonderPanda This will also require a bump

packages/modules/src/dynamicModules.ts
```diff
@@ -85,15 +85,22 @@ export interface IConfigurableDynamicRootModule<T, U> {
...
   externallyConfigured(
     moduleCtor: Type<T>,
-    wait: number
+    wait: number,
   ): Promise<DynamicModule>;
 }
 
+/**
+ *
+ * @deprecated Please use Nestjs Configurable Module {@see https://docs.nestjs.com/fundamentals/dynamic-modules#configurable-module-builder}
```
underfisk, line 99: I have added a deprecation/redirect to the nestjs module builder. We could potentially remove this module as it's not worth to be maintaining since Nestjs now has a better alternative

PR #928, opened 2025-01-12
feat: connection `wait` logging improvements and a new option to backfill the previous behaviour

### LINKED ISSUE #882: After updating the library, the application on NestJs stopped starting
Previously, in some of our projects, both in the dev environment and locally, we did not connect to real data but used random strings (in RMQ_USERNAME, RMQ_PASSWORD, RMQ_HOST). The application would start and display errors in the console. However, after the latest update (5.6.1), the application fails to start entirely. The configuration settings are provided below.
```javascript
export const RmqConfig: RabbitMQConfig = {
  uri: `amqp://${process.env.RMQ_USERNAME}:${process.env.RMQ_PASSWORD}@${process.env.RMQ_HOST}`,
  exchanges,
  queues,
  channels: {
    'main-channel': {
      prefetchCount: 1,
      default: true,
    },
  },
};
```
```javascript
@Module({
  imports: [RabbitMQModule.forRoot(RabbitMQModule, RmqConfig)],
  exports: [RabbitMQModule],
})
export class RmqModule {}
```

underfisk:
## Improvements
- **Connection Failure Logging**: When the wait option is set to true, connection failures will now log a warning message without a stack trace. This reduces visual clutter and avoids giving the impression that the application has crashed or encountered an unhandled error.
- `skipConnectionFailedLogging` : A new connectionOptions flag, skipConnectionFailedLogging, has been introduced. This allows developers to fully suppress connection failure logs, replicating the previous experience we had with amqplib.
- **Enhanced JSDoc**: The JSDoc for ConnectionInitOptions has been improved for better clarity and completeness, helping developers understand the behavior and usage of each configuration option.

Closes #882

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -109,6 +109,7 @@ const defaultConfig = {
     wait: true,
     timeout: 5000,
     reject: true,
+    skipConnectionFailedLogging: false,
   },
   connectionManagerOptions: {},
   registerHandlers: true,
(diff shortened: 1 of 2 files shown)


underfisk:
@WonderPanda We'll need to release this possibly with the major upgrade that we have to do due to the module definition API.

The new `amqlib` behaviour is more verbose but consumers leveraging `wait` do not wanna see this "false positive" messages when they are aware that the broker is possibly unavailable for a while.

PR #945, opened 2025-01-22
fix(discovery): support NestJS 11 

### LINKED ISSUE #944: NestJS 11 support
Renovate opened a [PR](https://github.com/golevelup/nestjs/pull/937) to support NestJS 11. 

Currently CI is failing, if I fix the issue will you create a new release to support NestJS 11?

xballoy:
Fix #937 to support NestJS 11.

Starting version 11, `InstanceWrapper.inject` can be `null` (see [v11](https://github.com/nestjs/nest/blob/v11.0.0/packages/core/injector/instance-wrapper.ts#L69) vs [v10](https://github.com/nestjs/nest/blob/v10.4.15/packages/core/injector/instance-wrapper.ts#L68)).

In order to prevent a breaking change in discovery by respecting the `DiscoveredClass` type we map `null` to `undefined`.

Fixes #944

### DIFF
--- a/packages/discovery/src/discovery.service.ts
+++ b/packages/discovery/src/discovery.service.ts
@@ -223,7 +223,7 @@ export class DiscoveryService {
     return {
       name: wrapper.name as string,
       instance: instanceHost.instance,
-      injectType: wrapper.metatype,
+      injectType: wrapper.metatype ?? undefined,
       dependencyType: get(instanceHost, 'instance.constructor'),
       parentModule: {
         name: nestModule.metatype.name,
--- a/packages/discovery/src/discovery.service.ts
+++ b/packages/discovery/src/discovery.service.ts
@@ -223,6 +223,7 @@ export class DiscoveryService {
     return {
       name: wrapper.name as string,
       instance: instanceHost.instance,
+      // TODO: remove nullish coalescing operator to return undefined when dropping NestJS 10 support
       injectType: wrapper.metatype ?? undefined,
       dependencyType: get(instanceHost, 'instance.constructor'),
       parentModule: {
--- a/integration/rabbitmq/package.json
+++ b/integration/rabbitmq/package.json
@@ -24,7 +24,7 @@
   "devDependencies": {
     "@golevelup/nestjs-rabbitmq": "workspace:^",
     "@golevelup/ts-jest": "workspace:^",
-    "@nestjs/testing": "^10.4.4",
+    "@nestjs/testing": "^11.0.0",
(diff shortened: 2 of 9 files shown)


xballoy:
I successfully locally run

```
pnpm build --concurrency=12 
pnpm test:ci
pnpm test:vitest:ci
```

I did not run the `pnpm test:ci:integration` because I don't want to setup a local RabbitMQ but this package is not impacted anyway.

packages/discovery/src/discovery.service.ts
```diff
@@ -223,7 +223,7 @@ export class DiscoveryService {
     return {
       name: wrapper.name as string,
       instance: instanceHost.instance,
-      injectType: wrapper.metatype,
+      injectType: wrapper.metatype ?? undefined,
```
underfisk, line 227: @xballoy Shall we add a todo here so that when V10 is removed, we can revert it to the expected `null` in the future?
Backfill is fine for now as we're going to support both v10 and v11 but if devs fully adopt v11, this could be a library-specific behavior

underfisk:
> I successfully locally run
> 
> ```
> pnpm build --concurrency=12 
> pnpm test:ci
> pnpm test:vitest:ci
> ```
> 
> I did not run the `pnpm test:ci:integration` because I don't want to setup a local RabbitMQ but this package is not impacted anyway.

@xballoy Not an issue, CI will run those checks for you.

packages/discovery/src/discovery.service.ts
xballoy: Yes, we could add a reminder here for if we drop NestJS 10 support.

But at the same time, I guess we could change `DiscoveredClass.injectType` to be able to be null, I think that most of the time people are handling null/undefined the same way so this wouldn't really be a breaking change (maybe on the types). I just wanted the change to be minimal as I wasn't sure about the backward compatibility you wanted.

I'm open to both solutions, just tell me what you prefer.

underfisk: I agree with you, we can add the reminder and also improve the injectType.
As long as we're clear about that breaking change, we're good. I'd like to prepare devs that we're working towards the Nestjs norm on that front.

xballoy: I changed the following:
- keep returning undefined instead of null but adding a TODO for when we'll remove NestJS 10
- I added null to `DiscoveredClass.injectType` types so we prepare support for NestJS 11 (also added a TODO to remove the undefined)

underfisk (approved):
lgtm thank you

PR #996, opened 2025-03-24
feat(ts-jest): add strict option to createMock

icholy:
When `strict` is enabled, calls to unstubbed methods throw instead of returning another mock

### DIFF
--- a/docs/testing/ts-jest.md
+++ b/docs/testing/ts-jest.md
@@ -119,3 +119,31 @@ Be aware that when providing your own mocks, if you asserting how many times you
 The above case shows how well the `createMock` utility can take in user provided values as well as returning **type safe** mocks that can easily be chained and modified as needed.
 
 For a few more examples on what can be done [the mock.spec](src/mocks.spec.ts) file has some really cool examples that show pretty well just what is doable with this utility.
+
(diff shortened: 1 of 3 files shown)


underfisk:
@icholy This looks handy! Are you able to contribute to the documentation by adding an example/explanation of strict?

icholy:
@underfisk yes, I just wanted to make sure you were open to the idea before spending time writing docs.

underfisk:
> @underfisk yes, I just wanted to make sure you were open to the idea before spending time writing docs.

All good on my end, it looks good for the use-case you're proposing. Since this is an opt-in functionality and you have provided tests, I feel confident about pushing for the change. Documenting it will just make it clear that its only available for ts-jest and explain it a little bit

icholy:
@underfisk ready for review.

underfisk (approved):

underfisk:
@icholy Thank you for this change! I appreciate the contribution

icholy:
@underfisk does this library have a release schedule?

underfisk:
It will have to be released by @WonderPanda 

I don't think we have a release schedule, it's more adhoc

PR #1003, opened 2025-04-21
chore(dependencies): updates dependencies

underfisk:
- Updates a few dev dependencies such as prettier and eslint
- Updates `nestjs` versions to `v11` removing support for v10
- Replaces `volta` with `mise`


underfisk:
@WonderPanda We may want to release a new version. Since we're removing support for nestjs v10 by encouraging the upgrade to v11, a major could make more sense

PR #1008, opened 2025-04-21
chore: remove zod

underfisk:
- Removes `zod` as we're only using it in a single place which feels unnecessary


packages/hasura/package.json
```diff
@@ -38,8 +38,7 @@
   "dependencies": {
     "@golevelup/nestjs-discovery": "workspace:^",
     "@hasura/metadata": "^1.0.2",
-    "js-yaml": "^4.1.0",
-    "zod": "^3.24.2"
```
underfisk, line 42: We can always bring it back when there's a potential usage. but for a single `z.enum` it's just an extra dependency to maintain

PR #1020, opened 2025-04-22
fix(rabbitmq): improve URI assertion

### LINKED ISSUE #1009: Wrong URI validation not allowing to pass query params for amqplib
We have this RegEx validation for RabbitMQ URI here:
https://github.com/golevelup/nestjs/blob/a57a7508be863b326e26175f688df7dd40a67fdf/packages/rabbitmq/src/amqp/utils.ts#L40-L52

It should allow to pass query params like, for example, `frameMax`.

underfisk:
- Fixes the RabbitMQ URI assertion to leverage node `URL` constructor and assert that the protocol is valid

Closes #1009

### DIFF
--- a/packages/rabbitmq/src/amqp/utils.ts
+++ b/packages/rabbitmq/src/amqp/utils.ts
@@ -29,7 +29,7 @@ export function matchesRoutingKey(
 }
 
 const rabbitMQRegex =
-  /^amqps?:\/\/(([^:]+):([^@]+)@)?([^:/]+)(:[0-9]+)?(\/.*)?$/;
+  /^amqps?:\/\/(?:[^@]+@)?[^:/?#]+(?::\d+)?(?:\/[^?#]*)?(?:\?.*)?$/;
 
 /**
  * Validates a rabbitmq uri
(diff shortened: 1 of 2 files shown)


underfisk:
@noghartt may i have your 👀  here?

packages/rabbitmq/src/amqp/utils.ts
```diff
@@ -29,7 +29,7 @@ export function matchesRoutingKey(
 }
 
 const rabbitMQRegex =
-  /^amqps?:\/\/(([^:]+):([^@]+)@)?([^:/]+)(:[0-9]+)?(\/.*)?$/;
+  /^amqps?:\/\/(?:[^@]+@)?[^:/?#]+(?::\d+)?(?:\/[^?#]*)?(?:\?.*)?$/;
```
noghartt: actually, I thought in doing it in a simple way, just parsing it as a `URL` instance.

We could do something like:
```ts
try {
  const uri = new URL(uri);
  return uri.protocol.startsWith('amqp') || uri.protocol.startsWith('amqps')
} catch (e) {
  // ...
}
```

IMO is less complex than handling RegEx per se, what do you think?

underfisk: Sure, looks like a good alternative. I will review your approach once I have some time free today

noghartt (approved):
LGTM

underfisk:
Thanks @noghartt 
When @WonderPanda performs the next release, this will be included

PR #1056, opened 2025-06-28
feat(rabbitmq): provide support for connecting using a config object

### LINKED ISSUE #931: Feature Request: Support for connecting to RabbitMQ using an object instead of a URL
Hello team 😄,

First off, thank you for this amazing package! It's been incredibly helpful in integrating RabbitMQ with NestJS applications, and I truly appreciate the effort and thought put into its development.

I would like to request a feature to **allow connecting to RabbitMQ using an object instead of a URL**. Currently, the package only supports connections via a URL string. However, in certain cases, it's more convenient to define those connection parameters (specially options such as heartbeat, etc.) as an object.

The underlying library, [amqplib](https://amqp-node.github.io/amqplib/channel_api.html#connect:~:text=Connecting%20with%20an%20object%20instead%20of%20a%20URL), already supports this functionality. This would simplify and enhance the flexibility of connection configurations and allow users to leverage all the options amqplib provides.

Thank you for considering this request!

danielshvarts1207:
closes #931


packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -176,10 +176,23 @@ export type RabbitMQHandlers = Record<
   string,
   MessageHandlerOptions | MessageHandlerOptions[]
 >;
+export type RabbitMQUriConfig = RabbitMQUriConfigObject | string;
+
+export interface RabbitMQUriConfigObject {
```
underfisk: @DanielShvarts2004 Does the amq lib offers the config object types? If so i'd prefer to simply import those

packages/rabbitmq/src/tests/rabbitmq.utils.spec.ts
```diff
@@ -58,42 +62,86 @@ describe(matchesRoutingKey.name, () => {
...
-          'superbawl://rabbitmq:rabbitmq@localhost:4444',
-        ]),
-      ).toThrowError();
-    });
+  it("should throw when host doesn't exist in uri objecct", () => {
+    expect(() =>
+      converUriConfigObjectsToUris({
+        host: undefined,
+      }),
+    ).toThrowError(Error("Configuration object must contain a 'host' key."));
```
underfisk: FYI no need to wrap in `Error`

underfisk (changes requested):
I have a question around the type, if amq offers the interface i'd rather import it

underfisk:
@DanielShvarts2004 Tests are failing btw

underfisk (approved):

underfisk:
@danielshvarts1207 are you able to fix the tests?

danielshvarts1207:
Yes of course I'll fix them

Issue #904, opened 2025-01-04
Feature Request: Support for `consumerTag` Configuration in RabbitMQ Module


arshadazaad3:
## **Description**
Currently, the RabbitMQ module in `@golevelup/nestjs` does not support configuring the `consumerTag` when setting up a queue consumer. This feature would be beneficial for applications that need fine-grained control over consumer identifiers, especially in environments where multiple consumers operate on the same queue.

---

## **Proposed Solution**
Add a `consumerTag` option to the `queueOptions` object in the RabbitMQ configuration. This tag should be passed to the underlying `amqplib` consumer configuration.

---

## **Example Usage**
Here’s how the configuration could look with the new feature:

```typescript
RabbitMQModule.forRootAsync(RabbitMQModule, {
  useFactory: () => ({
    uri: 'amqp://localhost:5672',
    exchanges: [{ name: 'exchange_name', type: 'topic' }],
    queues: [
      {
        name: 'queue_name',
        options: {
          consumerTag: 'custom_consumer_tag',
        },
      },
    ],
  }),
});
```

## **Benefits**
- Enhanced Consumer Management: Allows applications to track and manage consumers more effectively.
- Best Practices: Supports RabbitMQ’s best practices for consumer identification.
- Ease of Use: Reduces the need for manual setup and channel management when working with consumers.

## **Alternatives**
Developers can currently use the managedConnection to create channels manually, but this approach:

- Requires significant additional setup.

underfisk:
@arshadazaad3 Have you explored the usage of `queueOptions`? We already seem to support `consumerOptions` which allows you to provide a consumer tag.
The way you provide `consumerTag` is through our decorator `RabbitRPC` or `RabbitSubscribe` 

```typescript
  @RabbitRPC({
    routingKey: 'rpc',
    exchange: 'exchange1',
    queue: 'rpc',
    queueOptions: {
      consumerOptions: {
        consumerTag: 'test-consumer',
      },
    },
  })
  rpc(message: object) {
    return {
      echo: message,
    };
  }
```

arshadazaad3:
Thank you, @underfisk, for the clarification! I appreciate the pointer to consumerOptions within queueOptions and the example using RabbitRPC or RabbitSubscribe.

[...]

PR #1077, opened 2025-09-10
feat: support for consumerTag configuration in RabbitMQ

aweigor:
Related to https://github.com/golevelup/nestjs/issues/904.

Since the consumerTag global settings can be used in several places, I have placed the processing in factory method, and forwarded the updated object from factory method further into the call function.

I found that queue options to which the consumerTag configuration was supposed to be added, extends the internal types of amqplib, so I had to declare the configuration one level higher. The final configuration can be look like this:

```
  useFactory: () => ({
    uri: 'amqp://localhost:5672',
    exchanges: [{ name: 'exchange_name', type: 'topic' }],
    queues: [
      {
        name: 'queue_name',
        consumerTag: 'custom_consumer_tag',
        options: {
          // ... more options here
        },
      },
    ],
  }),
});
```

If local consumerTag settings are passed during consumer initialization, global settings are ignored. Added a couple of tests to check this.

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -36,7 +36,7 @@ import {
   MessageHandlerErrorBehavior,
 } from './errorBehaviors';
 import { Nack, RpcResponse, SubscribeResponse } from './handlerResponses';
-import { isNull, has, set, clone } from 'lodash';
+import { isNull, merge } from 'lodash';
 import { matchesRoutingKey } from './utils';
 
 const DIRECT_REPLY_QUEUE = 'amq.rabbitmq.reply-to';
@@ -503,33 +503,32 @@ export class AmqpConnection {
       msgOptions: MessageHandlerOptions,
     ) => Promise<string>,
   ): Promise<SubscriptionResult> {
-    // reassign consumer tag if it has been declared as default queue
-    // property and has not been provided in msgOptions argument
-    // see https://github.com/golevelup/nestjs/issues/904
-    const _options = clone(msgOptions);
-    const optionsConsumerPropertyPath = [
-      'queueOptions',
-      'consumerOptions',
-      'consumerTag',
-    ].join('.');
-    const consumerTagPropertyName = 'consumerTag';
-
-    const queueConfig = this.config.queues.find(
-      (q) => q.name === _options.queue,
-    );
-    const needToReassignConsumerTag =
-      has(queueConfig, consumerTagPropertyName) &&
-      !has(_options, optionsConsumerPropertyPath);
-
-    if (needToReassignConsumerTag) {
-      set(_options, optionsConsumerPropertyPath, queueConfig.consumerTag);
-    }
-    // reassign logic end
-
     return new Promise((res) => {
-      this.selectManagedChannel(_options?.queueOptions?.channel).addSetup(
+      // Use globally configured consumer tag.
+      // See https://github.com/golevelup/nestjs/issues/904
+
+      const queueConfig = this.config.queues.find(
+        (q) => q.name === msgOptions.queue,
+      );
+
+      const consumerTagConfig: Partial<MessageHandlerOptions> =
+        queueConfig?.consumerTag
+          ? {
+              queueOptions: {
+                consumerOptions: {
+                  consumerTag: queueConfig.consumerTag,
+                },
+              },
+            }
+          : {};
+
+      this.selectManagedChannel(msgOptions?.queueOptions?.channel).addSetup(
         async (channel: ConfirmChannel) => {
-          const consumerTag = await setupFunction(channel, _options);
+          const consumerTag = await setupFunction(
+            channel,
+            // msgOptions over global config
+            merge(consumerTagConfig, msgOptions),
+          );
           res({ consumerTag });
         },
       );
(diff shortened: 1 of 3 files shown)


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -498,12 +498,38 @@ export class AmqpConnection {
...
+    // reassign consumer tag if it has been declared as default queue
+    // property and has not been provided in msgOptions argument
+    // see https://github.com/golevelup/nestjs/issues/904
+    const _options = clone(msgOptions);
+    const optionsConsumerPropertyPath = [
+      'queueOptions',
+      'consumerOptions',
+      'consumerTag',
+    ].join('.');
+    const consumerTagPropertyName = 'consumerTag';
```
underfisk: can we extract this to a constant or some place outside this scope?

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -498,12 +498,38 @@ export class AmqpConnection {
...
+    setupFunction: (
+      channel: ConfirmChannel,
+      msgOptions: MessageHandlerOptions,
+    ) => Promise<string>,
   ): Promise<SubscriptionResult> {
+    // reassign consumer tag if it has been declared as default queue
+    // property and has not been provided in msgOptions argument
+    // see https://github.com/golevelup/nestjs/issues/904
+    const _options = clone(msgOptions);
+    const optionsConsumerPropertyPath = [
```
underfisk: I'm also curious here 🤔

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -22,6 +22,7 @@ export interface RabbitMQQueueConfig {
   exchange?: string;
   routingKey?: string | string[];
   bindQueueArguments?: any;
+  consumerTag?: string | undefined;
```
underfisk, line 33: Are you able to add the JSDOC with some explanation and possibly a backlink to the rmq lib?

underfisk (changes requested):
I've left some feedback around the structure

aweigor:
I've made some refactoring and added a small description for the property. Please feel free to comment.

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -470,7 +470,7 @@ export class AmqpConnection {
     originalHandlerName: string,
     consumeOptions?: ConsumeOptions,
   ): Promise<SubscriptionResult> {
-    return this.consumerFactory(msgOptions, (channel) =>
+    return this.consumerFactory(msgOptions, (channel, msgOptions) =>
```
underfisk: The `msgOptions` param seems to collide with the msgOptions that we initially have. Is this the parametized version? an override? Can we simply rename it something that states it a lot easier without conflicting with the existing `msgOptions` ?

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -486,7 +486,7 @@ export class AmqpConnection {
     msgOptions: MessageHandlerOptions,
     consumeOptions?: ConsumeOptions,
   ): Promise<SubscriptionResult> {
-    return this.consumerFactory(msgOptions, (channel) =>
+    return this.consumerFactory(msgOptions, (channel, msgOptions) =>
```
underfisk: Same observation, the type for the callback is fine, is just the naming of this param when we're unwrapping it as it collides with the scoped `msgOptions`

packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -498,12 +498,37 @@ export class AmqpConnection {
...
+              },
+            }
+          : {};
+
       this.selectManagedChannel(msgOptions?.queueOptions?.channel).addSetup(
         async (channel: ConfirmChannel) => {
-          const consumerTag = await setupFunction(channel);
+          const consumerTag = await setupFunction(
+            channel,
+            // msgOptions over global config
```
underfisk: This comment drives me a bit off, are we saying that we're providing an override to the global configuration here? If so could we simplify the comment and explain that we're merging the global/default tag configuration with the parametized msg option? 
I'm just playing devil advocate to make sure this is easier to maintain later on

packages/rabbitmq/src/rabbitmq.interfaces.ts
```diff
@@ -22,6 +22,15 @@ export interface RabbitMQQueueConfig {
   exchange?: string;
   routingKey?: string | string[];
   bindQueueArguments?: any;
+  /**
+   * This property can be set to define custom consumer name for the queue.
+   *
+   * Consumer tag must be unique to one channel. You shoudld properly manage
+   * channels and tags in your local space in order to avoid conflicts.
+   *
+   * More info: https://amqp-node.github.io/amqplib/channel_api.html#channel_consume
```
underfisk: Can you use JSDOC `@see` ?

underfisk (changes requested):
I'm feeling positive about this implementation, sorry for the late review.
There are some comments left around the variable naming that collide with the parameter

packages/rabbitmq/src/amqp/connection.ts
aweigor: The idea was to shadow msgOptions, because it saves us from having to ensure that the consumerFactory method does not change the original object. But yes, they may be different..

aweigor:
Added code that possibly fixes the issues. Oops about @ in commit message.

underfisk (approved):

PR #1082, opened 2025-11-16
feat(@golevelup/nestjs-google-cloud-pubsub): add new module

### LINKED ISSUE #1080: [Proposal] New Module: @golevelup/nestjs-google-cloud-pubsub
### Feature Request

This is a proposal to introduce a new module, `@golevelup/nestjs-gcp-pubsub`, to the `golevelup/nestjs` collection.

### Problem
Integrating GCP Pub/Sub with NestJS using the standard libraries involves a significant amount of manual setup and boilerplate.

Developers are left to manually manage:
* The configuration, validation, and mapping of **topics, subscriptions, schemas, and their revisions**.
* Custom **serialization/deserialization** logic for different schema types (like Avro or ProtoBuf).
* **Graceful shutdown** of subscriptions to ensure messages are not lost when the application terminates.

This creates a high risk of runtime errors due to configuration drift and adds significant, repetitive complexity.

### Proposed Solution
I propose a new module that solves this by **integrating a core validation client with a new, type-safe API layer**, all wrapped in a NestJS module.

The solution has two parts:

**1. The Core Client (Already Implemented):**
* A framework-agnostic `PubsubClient` that handles all critical *runtime/compile-time* logic: config drift detection, schema validation against remote revisions, and serialization.
* Exposes a `.close()` method to provide the *logic* for graceful shutdown.

**2. The Type-Safe NestJS Layer (Work in Progress):**
* This is the main part of the contribution, which I am **currently developing** (approx. 1 month of work remaining).
* It introduces the **`.initializeKit` method**, a type-inference engine that generates a fully-typed API (`Publisher`, `Subscribe`, `Payloads`) from a static config.
* It provides the **`NestjsGcpPubsubModule`** responsible for:
    * Calling the core client's validation logic on `onModuleInit`.
    * Hooking the core client's `.close()` method into `onModuleDestroy`.
    * Integrating the `.initializeKit` with NestJS's DI.

The initial implementation will focus on the **pull** model for subscriptions (the **push** model is out of scope for this initial version).

---

### Status

To be clear, this is a work in progress that is already significantly underway.

* **What is Done:** The framework-agnostic **core client** (handling runtime/compile-time validation, serialization, shutdown) is complete and can be reviewed.
* **What is In Progress:** I am currently building the **`.initializeKit` type-safety engine** and the **NestJS module** to integrate it (40% already ready).

I am creating this issue to get early feedback and, most importantly, **to understand if this is a feature you are interested in accepting** before I complete the final integration.

**The *completed core client* (which serves as the foundation) can be reviewed here:**
➡️ **https://github.com/chaikin-nikita/google-cloud-pubsub**

---

### Proposed API
The proposed API (which will be implemented by the new module) is designed to be highly type-safe, leveraging the `.initializeKit`.

*`pubsub.config.ts`*
```typescript
import { Encodings, SchemaTypes } from '@google-cloud/pubsub';
import { pubsubClient } from '@golevelup/nestjs-gcp-pubsub';

const { Publisher, Subscribe, topicConfigurations, Payloads } = pubsubClient.initializeKit([
  {
    name: 'order.created',
    schema: {
      definition: {
        fields: [
          { name: 'field1', type: 'string' },
          { name: 'field2', type: 'int' },
          { name: 'field3', type: 'boolean' },
        ],
        name: 'Level3',
        type: 'record',
      },
      encoding: Encodings.Json,
      name: 'order.created.schema',
      type: SchemaTypes.Avro,
    },
    subscriptions: [{ name: 'order.created.subscription.analytic-service' }],
  },
  // ... other topics
]);

export { Publisher, Subscribe, topicConfigurations, Payloads };
```

*`app.module.ts`*
```typescript
import { Module } from '@nestjs/common';
import { NestjsGcpPubsubModule, GcpPubsubMessage } from '@golevelup/nestjs-gcp-pubsub';
import { Publisher, topicConfigurations } from './pubsub.config';
import { AppService } from './app.service';

export const PUBLISHER_TOKEN = Symbol('TYPED_PUBLISHER');

@Module({
  imports: [
    NestjsGcpPubsubModule.registerAsync({
      topicConfigurations,
      Publisher: { injectionToken: PUBLISHER_TOKEN, ref: Publisher },
    }),
  ],
  providers: [AppService],
})
export class AppModule {}
```

*`app.service.ts`*
```typescript
import { Inject, Injectable } from '@nestjs/common';
import { GcpPubsubMessage } from '@golevelup/nestjs-gcp-pubsub';
import { Subscribe, Publisher, Payloads } from './pubsub.config';
import { PUBLISHER_TOKEN } from './app.module';

@Injectable()
export class AppService {
  constructor(
    @Inject(PUBLISHER_TOKEN) private readonly publisher: InstanceType<typeof Publisher>
  ) {}

  public async publishMethod() {
    // Autocomplete for topic name: 'order.created'
    // Autocomplete and type-checking for the payload object!
    await this.publisher.publish('order.created', {
      field1: 'hello',
      field2: 123,
      field3: true,
      // field4: 'abc' // <-- TypeScript Error!
    });
  }

  // Autocomplete for topic and subscription names
  @Subscribe('order.created', 'order.created.subscription.analytic-service')
  public async subscribeMethod(
    // Typescript will throw error because of wrong type
    payload: GcpPubsubMessage<boolean> 
  ) {}

  @Subscribe('order.created', 'order.created.subscription.analytic-service')
  public async validSubscribeMethod(
    payload: GcpPubsubMessage<Payloads['order.created']> 
  ) {}
}
```

chaikin-nikita:
Closes #1080


packages/google-cloud-pubsub/src/client/pubsub-configuration.errors.ts
underfisk, line 1: Could we either pluralize the file name or extract each error to its own file?

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
```diff
@@ -0,0 +1,166 @@
...
+} from './pubsub-configuration.error';
+import { PubsubTopicContainer } from './pubsub-topic.container';
+
+export class PubsubSchemaClient {
+  private _schemaClient: SchemaServiceClient | null = null;
+  private _schemaClientPromise: Promise<SchemaServiceClient> | null = null;
+
+  private readonly pubsub: PubSub;
+
+  constructor(pubsub: PubSub) {
```
underfisk: FYI you can do `private readonly` directly on the constructor which would create the private member for you

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.pubsub = pubsub;
+  }
+
+  private get schemaClient(): Promise<SchemaServiceClient> {
+    if (this._schemaClient) {
+      return Promise.resolve(this._schemaClient);
+    }
+
+    if (!this._schemaClientPromise) {
+      this._schemaClientPromise = this._initializeSchemaClient();
```
underfisk, line 27: Should we await the promise and make sure we return the resolved client?

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
```diff
@@ -0,0 +1,166 @@
...
+  private async _initializeSchemaClient(): Promise<SchemaServiceClient> {
+    this._schemaClient = await this.pubsub.getSchemaClient();
+
+    return this._schemaClient;
+  }
+
+  public async connectAndValidateSchema(topicContainer: PubsubTopicContainer) {
+    const schemaConfiguration = topicContainer.configuration.schema;
+
+    if (!schemaConfiguration) {
```
underfisk, line 42: Can we log warning that the configuration was not provided?

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
underfisk, line 1: Could we rename the directory to `e2e` and follow the existing convention? It would be a lot easier

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
```diff
@@ -0,0 +1,552 @@
...
+        name: 'Nested1',
+        type: 'record',
+      },
+    },
+  ],
+  name: 'Level3',
+  type: 'record',
+} satisfies schema.RecordType;
+
+// These tests are skipped. They require a real Google Cloud Pub/Sub instance.
```
underfisk, line 35: Are we able to set one up in CI? If you look at our RabbitMQ e2e, we do have real integration agains't a RMQ instance. I would love to see that here too

packages/google-cloud-pubsub/src/google-cloud-pubsub.constants.ts
```diff
@@ -0,0 +1,6 @@
+export const GOOGLE_CLOUD_PUBSUB_CLIENT_TOKEN = Symbol.for(
```
underfisk: Is there a reason why we would want a global symbol using `Symbol.for` instead of using `Symbol`?

packages/google-cloud-pubsub/src/google-cloud-pubsub.module-definition.ts
```diff
@@ -0,0 +1,25 @@
...
+  logger?: LoggerService;
+  topics: readonly PubsubTopicConfiguration[];
+};
+
+export const {
+  ConfigurableModuleClass,
+  MODULE_OPTIONS_TOKEN: GOOGLE_CLOUD_PUBSUB_MODULE_OPTIONS_TOKEN,
+} = new ConfigurableModuleBuilder<GoogleCloudPubsubModuleOptions>()
+  .setExtras(
+    {} as GoogleCloudPubsubModuleOptionsExtras,
```
underfisk: Would it be worth setting `global: true` as the baseline default?

package.json
```diff
@@ -83,5 +83,6 @@
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
-  }
+  },
+  "packageManager": "pnpm@10.19.0+sha512.c9fc7236e92adf5c8af42fd5bf1612df99c2ceb62f27047032f4720b33f8eacdde311865e91c411f2774f618d82f320808ecb51718bfa82c060c4ba7c76a32b8"
```
underfisk: This should be removed as we're leveraging `mise` for the pnpm/node versioning

underfisk (changes requested):
This is looking great, thanks for adding the library support!
I would love to see the e2e setup properly where we can set-up an e2e-app similar to our RMQ and run it in CI

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
chaikin-nikita: This is async singleton, we cannot await the promise here because we need to synchronously cache the pending execution.

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
chaikin-nikita: I think it's redundant here because the client is designed to work without a schema. It's normal behavior to work without schema

packages/google-cloud-pubsub/src/google-cloud-pubsub.constants.ts
chaikin-nikita: Good point, thanks. I'll fix that

packages/google-cloud-pubsub/src/google-cloud-pubsub.module-definition.ts
chaikin-nikita: Yeah, that makes sense. I'll make it global by default

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
underfisk: I'm not sure how i feel about introducing a singleton for this, I would rather have the client being a Nestjs Provider where you can either use "async" through `useFactory` or provide the object. This can proxied by our module to generate using the provisioned user configuration. 

Since we're talking about the client, a factory would help when the module is compiling to know whether it is possible to resolve or not whereas here, it would only tell you after the application run and the pubsub execute this block

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
chaikin-nikita: I thought about adding tests inside a container with the GCP Pubsub emulator, but it doesn't work exactly like the real cloud pubsub. It has issues with schema validation, and  especially with protobuf schemas(don't work at all).

I didn't want to add tests for the sake of tests, so i decided to skip it.
The tests provided here are fully functional. I run them locally against my personal GCP account.
Regarding real integration tests in CI: we can do this, but it requires a real Google Cloud Service Account. What do you think about it?

underfisk: Understood. I mean if we can setup a free google account managed by golevelup, I'm all in. The only reason why e2e tests would be valuable is to avoid regressions or even when doing "maintenance" like package upgrades to guarantee that things keep working as fine.

If you're able to help me out on what we would require (free if possible), better. I'm curious if they offer some Docker image that we could run.

chaikin-nikita: It will be so cool if you provide GCP account for me.
Regarding Docker images, yes, they have such one, but as I mentioned, this image does not support proto schema at all, and it's not a full copy of the real Cloud PubSub.

chaikin-nikita: I'll help you to set up e2e for pubsub.
Tomorrow, i’ll post a detailed set of instructions here on what needs to be configured on the Google Cloud side (API enablement, Service Account roles) and exactly what needs to be added to the GitHub Secrets to make the e2e working

underfisk: @chaikin-nikita In a first phase, I don't mind that we move these e2e tests to a directory and with a README, provide those instructions, even if we're not running them yet in CI but as a follow up, I would definetely wanna have something in CI as I doubt we'll likely remember to run it locally 😓

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
chaikin-nikita: It is strictly an internal helper, not a standalone service. Moving it to a nestjs provider would require exposing the internal PubSub instance to the DI container, which breaks encapsulation.
I think if we keep schemaClient internally, we maintain the decoupling of our client from NestJS, keeping the core logic framework-agnostic.
Also, regarding 'fail fast': if a schema is actually configured, the client attempts to resolve and validate it immediately during .initialize(). Since we await this method inside onApplicationBootstrap, any issues will throw an error right there and stop the application boot process.

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
chaikin-nikita: We can discuss how to run tests in ci later, it won't be a problem

packages/google-cloud-pubsub/src/client/pubsub-schema.client.ts
underfisk: Fair point!

packages/google-cloud-pubsub/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "build": "tsc --build tsconfig.build.json",
+    "build:watch": "tsc --build tsconfig.build.json --watch",
+    "test": "jest"
+  },
+  "bugs": {
+    "url": "https://github.com/golevelup/nestjs/issues"
+  },
+  "dependencies": {
+    "@golevelup/nestjs-discovery": "workspace:^",
+    "@google-cloud/pubsub": "^5.2.0"
```
underfisk: @chaikin-nikita Does it make more sense for this to be a peerDepedency? We can always install this as part of the devDependecies but it would give more control to people if they have this lib already installed

integration/google-cloud-pubsub/nodemon.json
underfisk, line 1: We don't need nodemon, we can simply leverage nest CLI to start

integration/google-cloud-pubsub/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "name": "google-cloud-pubsub-integration",
+  "version": "0.0.1",
+  "description": "Google cloud pubsub integration project",
+  "private": true,
+  "license": "MIT",
+  "scripts": {
+    "build": "tsc -p tsconfig.build.json",
+    "format": "prettier --write \"src/**/*.ts\" \"e2e/**/*.ts\"",
+    "start": "tsx src/main.ts",
+    "start:dev": "nodemon",
```
underfisk: You can start using Nest CLI 
https://docs.nestjs.com/cli/overview

underfisk:
@chaikin-nikita A few minor feedback, LMK when it  gets ready for review

chaikin-nikita:
> @chaikin-nikita A few minor feedback, LMK when it gets ready for review

@underfisk Done. Ready for review.

underfisk (approved):

chaikin-nikita:
@underfisk what do you think about adding something like this GCP PubSub, but for the Kafka based on @confluentinc/kafka-javascript ? Type safe, onBootstrap infrastructure validation(replicas, partitions, ...rest settings), underhood schema management.

underfisk:
> @underfisk what do you think about adding something like this GCP PubSub, but for the Kafka based on @confluentinc/kafka-javascript ? Type safe, onBootstrap infrastructure validation(replicas, partitions, ...rest settings), underhood schema management.

I'm not entirely sure as I don't have the time to maintain it though

chaikin-nikita:
> > @underfisk what do you think about adding something like this GCP PubSub, but for the Kafka based on @confluentinc/kafka-javascript ? Type safe, onBootstrap infrastructure validation(replicas, partitions, ...rest settings), underhood schema management.
> 
> I'm not entirely sure as I don't have the time to maintain it though

Okay, got it, anyway, i have enough work to do on this gcp pubsub, i'll focus on it

chaikin-nikita:
> > @underfisk what do you think about adding something like this GCP PubSub, but for the Kafka based on @confluentinc/kafka-javascript ? Type safe, onBootstrap infrastructure validation(replicas, partitions, ...rest settings), underhood schema management.
> 
> I'm not entirely sure as I don't have the time to maintain it though

@underfisk  Regarding maintaining, i can fully take care of its maintenance: issues, updates, and improvements, so it won't add extra work on your side

underfisk:
> > > @underfisk what do you think about adding something like this GCP PubSub, but for the Kafka based on @confluentinc/kafka-javascript ? Type safe, onBootstrap infrastructure validation(replicas, partitions, ...rest settings), underhood schema management.
> > 
> > 
> > I'm not entirely sure as I don't have the time to maintain it though
> 
> @underfisk Regarding maintaining, i can fully take care of its maintenance: issues, updates, and improvements, so it won't add extra work on your side

If you'd like to put a PoC/concept I'm down to see it but no guarantees :sweat_smile:

chaikin-nikita:
> > > > @underfisk what do you think about adding something like this GCP PubSub, but for the Kafka based on @confluentinc/kafka-javascript ? Type safe, onBootstrap infrastructure validation(replicas, partitions, ...rest settings), underhood schema management.
> > > 
> > > 
> > > I'm not entirely sure as I don't have the time to maintain it though
> > 
> > 
> > @underfisk Regarding maintaining, i can fully take care of its maintenance: issues, updates, and improvements, so it won't add extra work on your side
> 
> If you'd like to put a PoC/concept I'm down to see it but no guarantees 😅

Okay, okay :) Got it

chaikin-nikita:
@underfisk, I've pushed a fix for tests. Could you rerun tests in CI?

underfisk:
@chaikin-nikita As soon as I get the chance I'll try to see how we can get this published, thanks once again for getting it in.

chaikin-nikita:
@underfisk Great, thanks! Happy to help, just ping me if you need anything from my side(tests in ci, ...etc).

underfisk:
@chaikin-nikita The module has been published https://www.npmjs.com/package/@golevelup/nestjs-google-cloud-pubsub

Please let me know if this works/run into issues 🥳 
Docs were also added using your readme as the baseline, but feel free to modify them and improve if needed. https://golevelup.github.io/nestjs/modules/google-pubsub.html

chaikin-nikita:
@underfisk Wow, it's so cool, thanks.
I checked it, and everything works fine, but I just released a new version with some minor fixes. Could you republish it when you have a chance?

underfisk:
> @underfisk Wow, it's so cool, thanks. I checked it, and everything works fine, but I just released a new version with some minor fixes. Could you republish it when you have a chance?

Sure LMK when you have thge PR ready and i'll do it

chaikin-nikita:
@underfisk You can look at it here https://github.com/golevelup/nestjs/pull/1089/files

Issue #1083, opened 2025-11-19
have a dedicated error for rabbitmq rpc timeouts


Loki-Afro:
Could we please have a dedicated Exception/Error for when an RPC timeout occurs?

relevant code: https://github.com/golevelup/nestjs/blob/52e84ad3ddd9dfed9205bbe1d1dee565fb1a9931/packages/rabbitmq/src/amqp/connection.ts#L440-L447

Otherwise, you are forced to check for the message :/ 

Thank you!

underfisk:
@Loki-Afro I'm happy to accept contributions if you'd like to create a custom error. It should be a fairly simple chance but I don't really have the time now to add it

rkarbowiak:
@underfisk created smth like this here https://github.com/golevelup/nestjs/pull/1085

lmk what you think

Loki-Afro:
@rkarbowiak sorry didn't notice, thanks for merging it anyway!

Issue #686, opened 2024-01-26
[RabbitMQ] AmqpConnection.request() unhandled exception when losing connection


cs-salim:
I have nest microservices making rpc requests using `AmqpConnection.request()`, which crashes the whole microservice if the connection to RMQ is down at that moment.

Some errors are handled correctly but one is not, and it's just not possible to catch and handle it on my end:

These errors are handled, no problem here:

```
[my-microservice] [my-microservice] Error	2024-01-26T10:34:18-05:00 [AmqpConnection] Disconnected from RabbitMQ broker (default) - {
[my-microservice]   stack: [
[my-microservice]     'Error: Heartbeat timeout\n' +
[my-microservice]       '    at Heart.<anonymous> (/Users/me/repos/my-repo/node_modules/amqplib/lib/connection.js:427:19)\n' +
[my-microservice]       '    at Heart.emit (node:events:513:28)\n' +
[my-microservice]       '    at Heart.runHeartbeat (/Users/me/repos/my-repo/node_modules/amqplib/lib/heartbeat.js:88:17)\n' +
[my-microservice]       '    at listOnTimeout (node:internal/timers:559:17)\n' +
[my-microservice]       '    at processTimers (node:internal/timers:502:7)'
[my-microservice]   ]
[my-microservice] } +12s
[my-microservice] [my-microservice] Error	2024-01-26T10:34:18-05:00 [ExceptionsHandler] channel closed - {
[my-microservice]   stack: [
[my-microservice]     'Error: channel closed\n' +
[my-microservice]       '    at ConfirmChannel.<anonymous> (/Users/me/repos/my-repo/node_modules/amqplib/lib/channel.js:39:18)\n' +
[my-microservice]       '    at ConfirmChannel.emit (node:events:525:35)\n' +
[my-microservice]       '    at ConfirmChannel.C.toClosed (/Users/me/repos/my-repo/node_modules/amqplib/lib/channel.js:175:8)\n' +
[my-microservice]       '    at Connection.C._closeChannels (/Users/me/repos/my-repo/node_modules/amqplib/lib/connection.js:394:18)\n' +
[my-microservice]       '    at Connection.C.toClosed (/Users/me/repos/my-repo/node_modules/amqplib/lib/connection.js:401:8)\n' +
[my-microservice]       '    at Heart.<anonymous> (/Users/me/repos/my-repo/node_modules/amqplib/lib/connection.js:430:12)\n' +
[...]

PR #1084, opened 2025-11-19
Fix unhandled error

### LINKED ISSUE #1079: [RabbitMQ] Unhandled AggregateError when connection is lost during pending AmqpConnection.request() timeout
## Overview
When using AmqpConnection.request(), if the RabbitMQ connection is lost while a request is pending timeout, an unhandled AggregateError [ECONNREFUSED] crashes the application. This error occurs before the timeout error is thrown and cannot be caught with a standard try/catch block.​

## Steps to Reproduce
- Configure RabbitMQ module with valid connection settings
- Make a request using `amqp.request`with a timeout (e.g., 5000ms)
- While the request is waiting to timeout, close/stop the RabbitMQ broker
- Observe the unhandled `AggregateError [ECONNREFUSED]` that crashes the application

Let me know whether should I prepare a reproduction repository, but similar case is in this [issue](https://github.com/golevelup/nestjs/issues/686) already and it seems to have a reproduction repo.

## Expected Behavior
The connection loss error should be caught and handled gracefully, allowing the application to continue running. The timeout error should be thrown and catchable by the caller's try/catch block.​

## Actual Behavior
An unhandled AggregateError [ECONNREFUSED] is thrown, followed by the timeout error, crashing the entire application.

Error message:
```
AggregateError [ECONNREFUSED]: 
    at internalConnectMultiple (node:net:1139:18)
    at afterConnectMultiple (node:net:1714:7)
    at TCPConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17)
node_modules/@golevelup/nestjs-rabbitmq/lib/amqp/connection.js:212
            throw new Error(`Failed to receive response within timeout of ${timeout}ms for exchange "${requestOptions.exchange}" and routing key "${requestOptions.routingKey}"`);
                  ^

Error: Failed to receive response within timeout of 5000ms for exchange "test-exchange" and routing key "*"
    at <anonymous> (node_modules/@golevelup/nestjs-rabbitmq/src/amqp/connection.ts:431:15)
    at <anonymous> (node_modules/rxjs/src/internal/operators/map.ts:57:33)
    at OperatorSubscriber.OperatorSubscriber._this._next (node_modules/rxjs/src/internal/operators/OperatorSubscriber.ts:70:13)
    at OperatorSubscriber.Subscriber.next (node_modules/rxjs/src/internal/Subscriber.ts:71:12)
    at <anonymous> (node_modules/rxjs/src/internal/operators/throwIfEmpty.ts:50:22)
    at OperatorSubscriber.OperatorSubscriber._this._next (node_modules/rxjs/src/internal/operators/OperatorSubscriber.ts:70:13)
    at OperatorSubscriber.Subscriber.next (node_modules/rxjs/src/internal/Subscriber.ts:71:12)
    at <anonymous> (node_modules/rxjs/src/internal/operators/take.ts:60:26)
    at OperatorSubscriber.OperatorSubscriber._this._next (node_modules/rxjs/src/internal/operators/OperatorSubscriber.ts:70:13)
    at OperatorSubscriber.Subscriber.next (node_modules/rxjs/src/internal/Subscriber.ts:71:12)
```

## Environment

Node.js version: v23.11.0
@golevelup/nestjs-rabbitmq version: 6.0.1
NestJS version: 11.1.2

## Related issues

https://github.com/golevelup/nestjs/issues/259
https://github.com/golevelup/nestjs/issues/686

upundefined:
fixes https://github.com/golevelup/nestjs/issues/1079 https://github.com/golevelup/nestjs/issues/1083 https://github.com/golevelup/nestjs/issues/259 https://github.com/golevelup/nestjs/issues/686

### DIFF
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -445,7 +445,10 @@ export class AmqpConnection {
         );
       }),
     );
-    
+
+    // Wrapped lastValueFrom(race(response$, timeout$)) in a Promise to properly catch
+    // timeout errors. Without this, the timeout could trigger while publish() was
+    // still running, causing an unhandled rejection and crashing the application.
     const [result] = await Promise.all([
       lastValueFrom(race(response$, timeout$)),
       this.publish(
--- a/packages/rabbitmq/src/amqp/connection.ts
+++ b/packages/rabbitmq/src/amqp/connection.ts
@@ -445,22 +445,23 @@ export class AmqpConnection {
         );
       }),
     );
-
-    const result = lastValueFrom(race(response$, timeout$));
-
-    await this.publish(
-      requestOptions.exchange,
-      requestOptions.routingKey,
-      payload,
-      {
-        ...requestOptions.publishOptions,
-        replyTo: DIRECT_REPLY_QUEUE,
-        correlationId,
-        headers: requestOptions.headers,
-        expiration: requestOptions.expiration,
-      },
-    );
-
+    
+    const [result] = await Promise.all([
+      lastValueFrom(race(response$, timeout$)),
+      this.publish(
+        requestOptions.exchange,
+        requestOptions.routingKey,
+        payload,
+        {
+          ...requestOptions.publishOptions,
+          replyTo: DIRECT_REPLY_QUEUE,
+          correlationId,
+          headers: requestOptions.headers,
+          expiration: requestOptions.expiration,
+        },
+      ),
+    ]);
+    
     return result;
   }
 


packages/rabbitmq/src/amqp/connection.ts
```diff
@@ -445,22 +445,23 @@ export class AmqpConnection {
...
-        ...requestOptions.publishOptions,
-        replyTo: DIRECT_REPLY_QUEUE,
-        correlationId,
-        headers: requestOptions.headers,
-        expiration: requestOptions.expiration,
-      },
-    );
-
+    
+    const [result] = await Promise.all([
```
underfisk, line 452: @upundefined Are you able to add some sort of documentation/comment explaining the reasoning behind parallizing both calls?

underfisk (changes requested):
I'm not entirely sure about the new changes but also we would need test coverage in order to feel confident about the new changes/avoid regressions

packages/rabbitmq/src/amqp/connection.ts
upundefined: Explaining: previously, the code executed `const result = lastValueFrom(race(response$, timeout$));` without `await`, and then `await this.publish` was called. As a result, if `await this.publish` took longer than usual and the `timeout$` from `lastValueFrom(race(response$, timeout$))` triggered, an unhandled error occurred because there was no handler for it.

Now the line `const result = lastValueFrom(race(response$, timeout$))` is wrapped in a Promise, which allows the error to be caught without changing the logic of the previous code — the result from `lastValueFrom(race(response$, timeout$))` is still returned just as before.

upundefined:
> I'm not entirely sure about the new changes but also we would need test coverage in order to feel confident about the new changes/avoid regressions

I don't know how to do tests. But in my project, immediately after my fix, everything started working perfectly, there are no errors. I struggled with this error for a long time because it was crashing the entire application, which caused short-term outages. Eventually, I implemented a fix.

packages/rabbitmq/src/amqp/connection.ts
underfisk: Understood. Not entirely sure if the current implementation prior to your changes was worth already a restructuring to allow what you're describing. Although this makes sense, having a comment around this change in code will bring us some benefits for maintaining it later and avoid simply removing it without understanding its context

underfisk:
> > I'm not entirely sure about the new changes but also we would need test coverage in order to feel confident about the new changes/avoid regressions
> 
> I don't know how to do tests. But in my project, immediately after my fix, everything started working perfectly, there are no errors. I struggled with this error for a long time because it was crashing the entire application, which caused short-term outages. Eventually, I implemented a fix.

Tests would really help us catching your behavior without any potential regression. You can take a look at our e2e tests, it should be fairly simple to simulate what you're fixing here

packages/rabbitmq/src/amqp/connection.ts
upundefined: done

upundefined:
@underfisk i still don’t understand how to write a test where `this.publish` doesn’t respond, so that I can catch the `timeout$` error. If someone could help me with this, that would be great.

underfisk (approved):
Since this is a straight change and I doubt it would break any existing App out of the box, I'm okay on not having coverage yet on the issue as we're observing in order to unblock @upundefined 

I'll try to release a minor version with these changes as soon as I get the chance

underfisk:
@upundefined `nestjs-rabbitmq` package has been updated to `6.1.0`. Give it a spin and let me know

upundefined:
Works well for me, thank you!

PR #1085, opened 2025-11-28
feat(rabbitmq): adds several custom errors for better error handling

### LINKED ISSUE #1083: have a dedicated error for rabbitmq rpc timeouts

rkarbowiak:
### Changes

Adds custom error classes to the RabbitMQ module for better error handling and debugging.

Closes: #1083


packages/rabbitmq/src/amqp/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(
+    public readonly timeout: number,
+    public readonly exchange: string,
+    public readonly routingKey: string,
+  ) {
+    super(
+      `Failed to receive response within timeout of ${timeout}ms for exchange "${exchange}" and routing key "${routingKey}"`,
+    );
+    this.name = 'RpcTimeoutError';
+    Object.setPrototypeOf(this, RpcTimeoutError.prototype);
```
rkarbowiak: Due to [this](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work)

underfisk: What's the proposal fix? Is this due to the current typescript version we're running?

underfisk (approved):
LGTM thanks for contributing this!

rkarbowiak:
> LGTM thanks for contributing this!

Could you run CI to check if tests are passing?

packages/rabbitmq/src/amqp/errors.ts
rkarbowiak: It's more of a way to make sure it is compatible with older ecma versions e.g ES5, added it just to be safe, but can be deleted safely if only newer versions are supported by this library

underfisk: I see what u mean, i would suggest dropping as we want to incentivize modern ES though

rkarbowiak: Adjusted, would you run CI to check if it's passing? I see CI needs approval from maintainer

PR #1086, opened 2025-11-30
Fix changelog link for @golevelup/ts-jest

pwn-0x309:
Updated the changelog link for @golevelup/ts-jest package.


underfisk (approved):
Thank you!

PR #1089, opened 2025-12-08
feat(@golevelup/nestjs-google-cloud-pubsub): added handler for batch …

chaikin-nikita:
* Added batch handler
* Added sample for integration/google-cloud-pubsub
* Eslint config updated
* Serializing for topics without schema was removed(because it affects dlq topics)
* Abort controller replaced with explicit shutdown flow

### DIFF
--- a/integration/google-cloud-pubsub/.prettierrc
+++ b/integration/google-cloud-pubsub/.prettierrc
@@ -1,4 +1,11 @@
 {
+  "arrowParens": "always",
+  "bracketSpacing": true,
+  "endOfLine": "lf",
+  "printWidth": 155,
+  "semi": true,
   "singleQuote": true,
-  "trailingComma": "all"
+  "tabWidth": 2,
+  "trailingComma": "all",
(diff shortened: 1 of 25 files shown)


underfisk (approved):
lgtm, are you able to improve the documentation around the google pubsub?

Currently, we have the README on the project but usually the Docs readme, i copy-paste as I don't know an easy way to automate it. You can either simplify the readme on the project and treat the docs as the single source of truth as it is slightly better to work with.  
Can be a follow up for sure

underfisk:
Published under `0.0.2`

chaikin-nikita:
> lgtm, are you able to improve the documentation around the google pubsub?
> 
> Currently, we have the README on the project but usually the Docs readme, i copy-paste as I don't know an easy way to automate it. You can either simplify the readme on the project and treat the docs as the single source of truth as it is slightly better to work with. Can be a follow up for sure

Okay, for sure! I'll update the docs in a separate follow-up PR

PR #1087, opened 2025-12-03
fix: support stripe 20

arthurfiorette:
This PR updates nestjs-stripe package to support for Stripe's latest v20 release and also adds support for stripe's thin events

PR #1090, opened 2025-12-15
feat: add support for stripe's new v2 thin events

arthurfiorette:
This PR adds support for Stripe's v2 thin events via `@StripeThinWebhookHandler` decorator. 

Previously, with #1087 I've added a way to handle those, but now stripe requires a separate webhook for those, so I've created @StripeThinWebhookHandler and refactored a bit of the code to support both webhook types.

The webhook controller now accepts a `?mode=thin` query parameter to route thin events to their handlers, while snapshot events continue to work as before with no breaking changes.

Tests and docs were also updated.

https://docs.stripe.com/event-destinations#events-overview


arthurfiorette:
While this PR isn't merged/released, I'm releasing `@arthurfiorette/golevelup-nestjs-stripe`. Once it gets released, I'll deprecate that package.

Nsttt (approved):

packages/stripe/src/stripe.module.ts
```diff
@@ -113,50 +134,116 @@ export class StripeModule
...
+      return false;
+    }
+
+    return true;
+  }
+
+  public async onModuleInit() {
+    // If they didn't provide a webhook config secret there's no reason
+    // to even attempt discovery
+    if (!this.stripeModuleConfig.webhookConfig) {
```
underfisk, line 193: Shall we warn the developer instead? This should probably not even be possible

packages/stripe/src/stripe.webhook.controller.ts
```diff
@@ -21,15 +21,25 @@ export class StripeWebhookController {
...
-    @Request() request
+    @Request() request,
+    @Query('mode') mode: StripeWebhookMode = StripeWebhookMode.SNAPSHOT,
   ) {
     if (!sig) {
       throw new Error('Missing stripe-signature header');
     }
+
+    if (mode !== 'thin' && mode !== 'snapshot') {
+      throw new Error('Invalid mode query parameter');
```
underfisk: Could you include the provided `mode` interpolated in the error? That will help when debugging

underfisk (changes requested):
Small suggestions but LGTM overall

packages/stripe/src/stripe.module.ts
arthurfiorette: Some users might not use webhooks, and use this module just to get the stripe client.

packages/stripe/src/stripe.webhook.controller.ts
arthurfiorette:
```suggestion
      throw new Error(`Invalid mode ${mode} query parameter`);
```

arthurfiorette: done

packages/stripe/src/stripe.module.ts
underfisk: Ah i see, I haven't used it much before so I wasn't aware. We're good then!

underfisk (approved):

underfisk:
Available in the new version [0.11.0 ](https://www.npmjs.com/package/@golevelup/nestjs-stripe)
Thanks for the feature, LMK if there's any issue

PR #1091, opened 2025-12-18
feat: Graphile Worker support

underfisk:
### Improvements
- This PR introduces a Nestjs module for [Graphile Worker](https://worker.graphile.org/) 
   - ESM-only library, it's worth noting that golevelup libraries are going to adopt it even further!
- `@golevelup/nestjs-graphile-worker` brings some production-ready features such as runtime validations through Zod (in the future we may adopt the standardschema spec)
- Users can define their custom tasks to have type-safe autocompletion through Augmentation. 
More information available in the documentation section at our docs page

### Refactors
- Removed duplicated documentation from the new google-pubsub module
- Upgraded Typescript on our monorepo to the latest version `5.9.3` 
- Removed CI support for node 20 & 22 and added node 24 (LTS) to the matrix
- Updated CI PNPM version to v11


packages/google-cloud-pubsub/readme.md
underfisk, line 1: Centralizing documentation in our `docs` directory is preferable

packages/graphile-worker/vite.config.mjs
underfisk, line 1: As a follow-up, I will be migrating our tests over to Vitest and hopefully deprecate Jest as our internal test runner

PR #1113, opened 2025-12-28
feat(@golevelup/nestjs-google-cloud-pubsub): batching, async initialization, migrating docs

chaikin-nikita:
- Refactored initialization to be fully asynchronous, removing synchronous loops.
- Added decorator GoogleCloudPubsubBatchSubscribe.
- Migrated docs to docs/modules/google-pubsub.md as a single source of truth.


integration/google-cloud-pubsub/e2e/pubsub-client-connect-and-validate-topic.e2e-spec.ts
```diff
@@ -1,10 +1,7 @@
 import { PubSub } from '@google-cloud/pubsub';
 import * as crypto from 'crypto';
 
-import {
-  PubsubConfigurationInvalidError,
-  PubsubConfigurationMismatchError,
-} from '../../../packages/google-cloud-pubsub/src/client/pubsub-configuration.errors';
+import { PubsubConfigurationMismatchError } from '../../../packages/google-cloud-pubsub/src/client/pubsub-configuration.errors';
```
underfisk: @chaikin-nikita Instead of importing from packages (noticed it now), we should install the package as part of the integration deps. 
To get the same experience, all you will have to do is specify in the range `workspace:*` (you can look at the RMQ one)

packages/google-cloud-pubsub/src/client/pubsub-configuration.errors.ts
```diff
@@ -37,3 +37,13 @@ export class PubsubConfigurationInvalidError extends Error {
     this.name = PubsubConfigurationInvalidError.name;
   }
 }
+
+export class PubsubBatchManagerConfigurationInvalidError extends Error {
```
underfisk, line 41: Could we rename it to `PubSubBatchManagerInvalidConfigurationError`? I think it reads better

packages/google-cloud-pubsub/src/client/pubsub.client-types.ts
```diff
@@ -36,32 +29,23 @@ export interface PubsubClientConfiguration extends ClientConfig {
...
             : never
           : Buffer
       : Buffer
     : Buffer;
 };
 
-const invariantDataSymbol = Symbol.for(
-  'GOOGLE_CLOUD_PUBSUB_MESSAGE_INVARIANT_DATA',
-);
+const invariantDataSymbol = Symbol.for('GOOGLE_CLOUD_PUBSUB_MESSAGE_INVARIANT_DATA');
```
underfisk: Could we use instead just `Symbol` ?

packages/google-cloud-pubsub/src/google-cloud-pubsub.module-definition.ts
```diff
@@ -21,12 +21,9 @@ export const {
...
-    { isGlobal: true } as GoogleCloudPubsubModuleOptionsExtras,
-    (definition, extras) => ({
-      ...definition,
-      global: extras.isGlobal,
-      ...extras,
-    }),
-  )
+  .setExtras({ isGlobal: true } as GoogleCloudPubsubModuleOptionsExtras, (definition, extras) => ({
+    ...definition,
+    global: extras.isGlobal,
```
underfisk: This is likely redudant because the spread at the next line will override this key

packages/google-cloud-pubsub/.prettierrc
underfisk, line 1: Let's use the workspace prettier and remove this file

packages/google-cloud-pubsub/README.md
underfisk, line 1: Let's remove the readme from here, we've got a CI step to copy-paste from the docs.
The single source at the `docs` is what we truly need

underfisk (changes requested):
A couple notes but LGTM, thanks for adding this new additions

I'm not sure if you're developing with the whole monorepo but tool configurations like Node version/pnpm are configured through `mise` and `prettier` is also defined at the root

chaikin-nikita:
@underfisk I'll fix everything that you've mentioned in 2-3 days. I'm very busy right now

packages/google-cloud-pubsub/src/client/pubsub-configuration.errors.ts
chaikin-nikita: I'd prefer to keep it as is for consistency with the existing errors in this file (PubsubConfigurationMismatchError, PubsubConfigurationInvalidError)

chaikin-nikita:
@underfisk, PR updated - could you review again?

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
```diff
@@ -4,7 +4,11 @@ import * as crypto from 'crypto';
 import * as fs from 'fs';
 import * as path from 'path';
 
-import { GoogleCloudPubsubMessage, PubsubClient, PubsubTopicConfiguration } from '../../../packages/google-cloud-pubsub/src/client';
+import {
+  GoogleCloudPubsubMessage,
+  PubsubTopicConfiguration,
+} from '@golevelup/nestjs-google-cloud-pubsub/src';
+import { PubsubClient } from '@golevelup/nestjs-google-cloud-pubsub/src/client/pubsub.client';
```
underfisk: @chaikin-nikita Could you update the `index.ts` in the library so that you never have to reference the `src`?
Ideally the import should be only `@golevelup/nestjs-google-cloud-pubsub`

underfisk (changes requested):
Let's export the classes such as `PubSubClient`, errors etc that you'll need to import otherwise we're referencing absolute directory paths within the npm package 😅

packages/google-cloud-pubsub/src/index.ts
underfisk, line 1: @chaikin-nikita Is the `src/index` the main entry file for this library? If so we shall review how the imports are being made as we shouldn't be doing `@golevelup/library-name/src`, it should be the root itself

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
chaikin-nikita: @underfisk  PubsubClient is designed to be an internal implementation detail and errors also, these components cannot be accessible from the developer side.  I intentionally didn't export it in index.ts to keep the public API surface minimal and clean.

The deep import here is used strictly for 'white-box' integration testing to verify internal state. I believe it's safer to keep the path "ugly" in the tests than to expose internal parts of the library just for the sake of beautiful test imports. What do you think?

chaikin-nikita: I can move these tests to packages/google-cloud-pubsub/e2e if you want

packages/google-cloud-pubsub/src/index.ts
chaikin-nikita: Yes, src/index.ts is indeed the main entry point, and it already exports everything a developer needs for standard usage.

integration/google-cloud-pubsub/e2e/pubsub-client-attach-handler.e2e-spec.ts
underfisk: I agree with not exposing internals if thats the case here. Possibly having the e2e then in the library itself would be a lot easier.
What are your thoughts on having these e2e/integration tests in a separate project (currently) or moving them to the same as the lib?

chaikin-nikita: @underfisk, I moved tests to the lib

underfisk (approved):

underfisk:
@chaikin-nikita The remaining part is making sure that both proto files and e2e tests are not included in the tsconfig? Are you able to double check that?
If we do `tsc` to produce a build, those files must never be bundled

chaikin-nikita:
> @chaikin-nikita The remaining part is making sure that both proto files and e2e tests are not included in the tsconfig? Are you able to double check that? If we do `tsc` to produce a build, those files must never be bundled

I tested it and ran tsc from packages/google-cloud-pubsub, everything is okay, e2e and proto were not included to the build.

underfisk:
@chaikin-nikita https://github.com/golevelup/nestjs/pull/1119 Please verify in the new version if everything is good and LMK

chaikin-nikita:
> @chaikin-nikita #1119 Please verify in the new version if everything is good and LMK

@underfisk I've checked, everything is good

PR #1114, opened 2025-12-30
fix(ts-vitest): correct DeepMocked type definitions

icholy:
This PR fixes incorrect type definitions in `@golevelup/ts-vitest` that caused `mockResolvedValueOnce` and similar methods to have `never` types, and adds a typecheck CI step to catch type errors in test files.

## Type Fix

The `DeepMockedFunction` type was using `Mock<Parameters<T>>` but Vitest's `Mock` takes the full function type, so it should be `Mock<T>`. Additionally, `DeepMocked<T>` was missing the `& T` intersection that exists in the ts-jest version, which prevented mocks from being assignable to original types when private fields were present.

## CI

While adding a regression test for this fix, I discovered that the existing test suite already exercises these types extensively. Running `tsc` on the spec files revealed 7 type errors that were going unnoticed because `tsconfig.build.json` excludes `*.spec.ts`. Rather than adding another test, enabling type checking for test files in CI made more sense. This PR adds a `typecheck` script to all packages and a "Lint & Typecheck" step to the CI workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Build
         run: pnpm build --concurrency=12
 
+      - name: Lint & Typecheck
+        run: pnpm lint && pnpm lerna:typecheck
+
       - name: Run unit tests
         run: pnpm test:ci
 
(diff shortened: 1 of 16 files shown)


underfisk (approved):
Thanks for the fix and the CI addition!

icholy:
@underfisk what's your usual release cadence?

underfisk:
> @underfisk what's your usual release cadence?

I will try to get it release early this week

underfisk:
@icholy New version available #1118 for more details

icholy:
thanks!

PR #1115, opened 2025-12-30
test(ts-vitest): add strict mode test

icholy:
Add strict mode test for `createMock` and docs that mirror the ts-jest package.


underfisk (approved):

underfisk:
Thank you @icholy

PR #1117, opened 2026-01-04
ts-jest regression in 1.0.0 due to DeepMocked type changes

rbnayax:
Recent changes in https://github.com/golevelup/nestjs/pull/1075 is causing issued with complex types:
```
error TS2589: Type instantiation is excessively deep and possibly infinite
error TS2590: Expression produces a union type that is too complex to represent
```
This PR fixes that (based on the typings of `ts-vitest`)
I couldn't find a way to reproduce it without installing the aws sdk

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,6 @@ jobs:
         uses: pnpm/action-setup@v4
         with:
           version: 10
-          cache: true
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -38,7 +38,6 @@ jobs:
         uses: pnpm/action-setup@v4
         with:
           version: 10
-          cache: true
```
underfisk, line 41: Is there a particular reason to have the pnpm cache removed?

underfisk (approved):

underfisk:
@rbnayax Thank you, i will have it published with the ts-vitest this week

.github/workflows/ci.yml
rbnayax: see warnings in https://github.com/golevelup/nestjs/actions/runs/20693101548

underfisk:
@rbnayax New version available #1118 for more details

PR #1123, opened 2026-01-30
docs(stripe): fix for wrong import in example

NickTsitlakidis:
Minor fix for an incorrect import in the stripe readme.

### DIFF
--- a/docs/modules/stripe.md
+++ b/docs/modules/stripe.md
@@ -221,7 +221,7 @@ This library is built using an underlying NestJS concept called `External Contex
 You can identify Stripe webhook contexts by their context type, `'stripe_webhook'`:
 
 ```typescript
-import { STRIPE_WEBHOOK_CONTEXT_TYPE } from '@golevelup/stripe';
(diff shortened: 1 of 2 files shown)


underfisk (approved):

underfisk:
@NickTsitlakidis Thanks! Could you also verify if the docs for stripe also align? We're trying to migrate these package readmes to our doc has it brings better searchability

NickTsitlakidis:
@underfisk I've been using the package during the last week and I didn't notice any misalignment besides the one I fixed. is there a particular area in docs for which you're worried about?

underfisk:
> @underfisk I've been using the package during the last week and I didn't notice any misalignment besides the one I fixed. is there a particular area in docs for which you're worried about?

If you look at https://github.com/golevelup/nestjs/blob/master/docs/modules/stripe.md, we've got a different source for the documentation that gets generated by Vitepress. At some point, I may remove these duplicated READMEs because during build time, we copy the corresponding doc from that `docs` directory to the bundle, for NPM.
It's just so that if you're updating this file, we update the documentation one aswell

NickTsitlakidis:
Oh, I didn't know about the docs directory. The same issue existed there. I've fixed it.

PR #1262, opened 2026-09-23
feat: support NestJS 12

### LINKED ISSUE #1259: NestJS 12 support
NestJS 12 is finally out.

Does this library plan to support it?

As of now, we need to add `legacy-peer-deps=true` to .npmrc to make this work, otherwise npm install would fail because of peer dependencies.

For instance, @golevelup/nestjs-rabbitmq@9.0.2 (and its dependency @golevelup/nestjs-discovery) explicitly specifies peer dependencies restricted to NestJS 11 (@nestjs/common@^11.x and @nestjs/core@^11.x).

Solution: @golevelup/nestjs-rabbitmq publishes a release supporting NestJS 12 (e.g., peer dependency ^11.0.0 || ^12.0.0 or a new major version)

rodrigobnogueira:
Closes #1259.

* Widens the `@nestjs/common` / `@nestjs/core` peer ranges to `^11.1.24 || ^12.0.0` in every package that declares them (common, discovery, google-cloud-pubsub, hasura, modules, rabbitmq). The workspace stays on NestJS 11, so the existing CI keeps testing that end; the lockfile is unchanged.
* NestJS 12 is ESM-only and its exports map does not resolve directory imports. The published `.d.ts` of `nestjs-common`, `nestjs-modules` and `nestjs-webhooks` imported `@nestjs/common/interfaces`, which fails with `TS2307` for consumers on `node16`/`nodenext`/`bundler` resolution. Those imports now come from the `@nestjs/common` root, or from `@nestjs/common/interfaces/index.js` for the two types the root doesn't export (`RouteInfo`, `HttpArgumentsHost`) — the form `ts-vitest`'s spec already used. Both forms resolve on 11 and 12.
* One changeset: minor for the six widened packages, patch for webhooks.

What CI here can't show, since it runs NestJS 11: with every `@nestjs/*` forced to 12.0.4 (pnpm overrides), build, typecheck and the Vitest suite (111) pass, a `nodenext` consumer with `skipLibCheck: false` typechecks the published declarations cleanly (3× `TS2307` before), and a real-broker smoke test of `nestjs-rabbitmq` passes subscribe, RPC, an interceptor and `@RabbitPayload`. The Jest suites (hasura, rabbitmq e2e, ts-jest/ts-sinon specs) cannot load ESM-only NestJS 12 at all (`Unexpected token 'export'`), so moving the workspace itself to 12 will first need those moved to Vitest.

The pre-commit hook reformatted three of the touched files with oxfmt.


underfisk:
@rodrigobnogueira Thanks for putting this together. Can you take a look at the failing CI? once that's addressed i can get this merged

